PR #1, opened 2022-01-20
Add basic app structure with submodules

norling:
I was trying to make the structure simple to work with, so I used main only for help and argument parsing, then moved the "work" (currently just placeholders) into separate modules.

About command line args. I decided to keep things minimal, so there are very few args. This helps in making things simpler and also makes it very hard to do anything wrong.

Took me some time to get into golang mode again - so please comment if something's ugly or un-go-like :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,8 @@
+
+
+# personal files
+.vscode/
+personal/
+
+# compiled binary
+sda-cli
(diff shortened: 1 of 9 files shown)


norling:
Yes! :D I should really know that by now...

jbygdell (approved):

dbampalikis (approved):

PR #2, opened 2022-01-21
Add encryption module

norling:
Adds functionality to the encryption module.

I don't know if this code could be formatted better - but for now it works at least!
I'll make a separate PR with tests+CI, but I didn't want this PR to become too large.


encrypt/encrypt.go
```diff
@@ -27,14 +36,271 @@ var ArgHelp = `
...
+		return nil, err
+	}
+
+	// unencrypted sha256
+	shaHash := sha256.New()
+	_, err = io.Copy(shaHash, inFile)
+	if err != nil {
+		return nil, err
+	}
+	hashes.unencryptedSha256 = hex.EncodeToString(shaHash.Sum(nil))
```
jbygdell, line 208: This can be reworked a bit using the `TeeReader` or `MultiReader`, then you only need to read the file once while calculating both checksums.

encrypt/encrypt.go
```diff
@@ -27,14 +36,271 @@ var ArgHelp = `
...
+			outFilename = path.Join(*outDir, basename) + ".c4gh"
+		}
+
+		files = append(files, encryptionFileSet{filename, outFilename})
+	}
+
+	// Check that all the infiles exist, and all the outfiles don't
+	err := checkFiles(files)
+	if err != nil {
+		log.Fatal(err)
```
jbygdell, line 81: You are pulling in Logrus, so make use of it and generate proper verbose log messages

encrypt/encrypt.go
```diff
@@ -27,14 +36,271 @@ var ArgHelp = `
...
+func encrypt(filename, outFilename string, pubKey, privateKey [32]byte) error {
+	// check if outfile exists
+	if _, err := os.Stat(outFilename); err == nil {
+		return fmt.Errorf("outfile %s already exists", outFilename)
+	}
+
+	// read infile
+	inFile, err := os.Open(filename)
+	if err != nil {
+		return err
```
jbygdell, line 279: this kinds of errors are really annoying since you as a user don't get any information on `which` file it was that could not be opened.

jbygdell (changes requested):
1# tests are missing
2# some parts can be simplified
3# general comments

encrypt/encrypt.go
norling: Ah! :) I honestly thought the default error would include that. I'll make the error messages better!

encrypt/encrypt.go
norling: yes!

encrypt/encrypt.go
jbygdell: Sometimes it does, but when it doesn't it is really annoying.

jbygdell: Yet another reason to write the tests at the same time as the funcitons. ;)

encrypt/encrypt.go
```diff
@@ -27,14 +36,271 @@ var ArgHelp = `
...
+	}
+	publicKey, err := keys.ReadPublicKey(file)
+	if err != nil {
+		return nil, err
+	}
+	return &publicKey, nil
+}
+
+// Generates a crypt4gh key pair, returning only the private key, as the
+// public key isn't needed for encryption.
```
aaperis, line 257:
```suggestion
// public key used for encryption is the one provided as argument.
```
seemed kinda confusing, since a pk is needed for encryption but not that one

encrypt/encrypt.go
```diff
@@ -27,14 +36,271 @@ var ArgHelp = `
...
+
+	_, privateKey, err := keys.GenerateKeyPair()
+	if err != nil {
+		return nil, err
+	}
+	return &privateKey, nil
+}
+
+// Encrypts the data from `filename` into `outFilename` for the given `pubKey`,
+// using the given `privateKey`.
```
aaperis, line 269:
```suggestion
// using the generated `privateKey`.
```

please discard if I misunderstand sth

encrypt/encrypt.go
```diff
@@ -14,11 +23,11 @@ USAGE: %s encrypt [file(s)]
```
aaperis, line 23: Seems that line 22 should be 
 `USAGE: %s encrypt -key <public-key-file> [file(s)]`

aaperis (changes requested):
in addition to prev comments, there are a couple of things regarding output messages that would be nice to fix. One concerns the `main.go` where changing line 101 to sth like 
 `fmt.Fprintf(os.Stderr, info.usage, os.Args[0])` 
will correct the output formatting when no args are given. Otherwise, it looks good :-)

encrypt/encrypt.go
```diff
@@ -27,14 +36,288 @@ var ArgHelp = `
...
+
+	// Generate a random private key to encrypt the data
+	privateKey, err := generatePrivateKey()
+	if err != nil {
+		return err
+	}
+
+	// Open all checksum files
+	ChecksumFileUnencMd5, err := os.OpenFile("checksum_unencrypted.md5",
+		os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
```
jbygdell, line 86: There is no need to break this line

encrypt/encrypt.go
```diff
@@ -27,14 +36,288 @@ var ArgHelp = `
...
+	encryptedMd5      string
+	unencryptedMd5    string
+	encryptedSha256   string
+	unencryptedSha256 string
+}
+
+// Checks if a file exists in the file system. Note that this function will not
+// check if the file is readable, or if the file is a directory, only if it
+// exists.
+func FileExists(filename string) bool {
```
jbygdell, line 302:
```suggestion
func fileExists(filename string) bool {
```

Function names starting with a capital letter gets exported, no need since it is only used in this module

encrypt/encrypt.go
```diff
@@ -27,14 +36,288 @@ var ArgHelp = `
...
+	_, err := os.Stat(filename)
+	return err == nil
+}
+
+// Checks that a file exists, and is readable
+func FileIsReadable(filename string) bool {
```
jbygdell, line 308:
```suggestion
func fileIsReadable(filename string) bool {
```

Function names starting with a capital letter gets exported, no need since it is only used in this module

encrypt/encrypt.go
```diff
@@ -27,14 +36,288 @@ var ArgHelp = `
...
+		os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
+		return err
+	}
+	defer ChecksumFileEncSha256.Close()
+
+	// encrypt the input files
+	num_files := len(files)
+	for i, file := range files {
+		log.Infof("Encrypting file %v/%v: %s", i+1, num_files, file.unencrypted)
```
jbygdell, line 116:
```suggestion
	numFiles := len(files)
	for i, file := range files {
		log.Infof("Encrypting file %v/%v: %s", i+1, numFiles, file.unencrypted)
```

Don't use underscores in names

encrypt/encrypt.go
```diff
@@ -27,14 +36,288 @@ var ArgHelp = `
...
+
+		// Write hashes
+		ChecksumFileUnencMd5.WriteString(
+			fmt.Sprintf("%s %s\n", hashes.unencryptedMd5, file.unencrypted))
+		ChecksumFileUnencSha256.WriteString(
+			fmt.Sprintf("%s %s\n", hashes.unencryptedSha256, file.unencrypted))
+		ChecksumFileEncMd5.WriteString(
+			fmt.Sprintf("%s %s\n", hashes.encryptedMd5, file.encrypted))
+		ChecksumFileEncSha256.WriteString(
+			fmt.Sprintf("%s %s\n", hashes.encryptedSha256, file.encrypted))
```
jbygdell, line 137:
```suggestion
		// Write hashes
		if _, err := ChecksumFileUnencMd5.WriteString(fmt.Sprintf("%s %s\n", hashes.unencryptedMd5, file.unencrypted)); err != nil {
			return err
		}

		if _, err := ChecksumFileUnencSha256.WriteString(fmt.Sprintf("%s %s\n", hashes.unencryptedSha256, file.unencrypted)); err != nil {
			return err
		}

		if _, err := ChecksumFileEncMd5.WriteString(fmt.Sprintf("%s %s\n", hashes.encryptedMd5, file.encrypted)); err != nil {
			return err
		}

		if _, err := ChecksumFileEncSha256.WriteString(fmt.Sprintf("%s %s\n", hashes.encryptedSha256, file.encrypted)); err != nil {
			return err
		}
```

Always remember to check for errors

encrypt/encrypt.go
```diff
@@ -27,14 +36,288 @@ var ArgHelp = `
 
 var Args = flag.NewFlagSet("encrypt", flag.ExitOnError)
 
+var publicKeyFile = Args.String("key", "",
+	"Public key to use for encrypting files.")
+var outDir = Args.String("outdir", "", "Output directory for encrypted files")
+
 // Main encryption function
-func Encrypt(args []string) {
+func Encrypt(args []string) error {
```
jbygdell, line 44: comment should be in the form of:
```
// Encrypt is the main encryption function
```

decrypt/decrypt.go
```diff
@@ -25,7 +25,7 @@ var privateKey = Args.String("key", "",
 	"Private key to use for decrypting files.")
 
 // Main decryption function
-func Decrypt(args []string) {
+func Decrypt(args []string) error {
```
jbygdell, line 28: Should this change be here? it is not related to encryption

download/download.go
```diff
@@ -24,7 +24,7 @@ var ArgHelp = `
 var Args = flag.NewFlagSet("download", flag.ExitOnError)
 
 // Main download function
-func Download(args []string) {
+func Download(args []string) error {
```
jbygdell, line 27: Should this change be here? it is not related to encryption

main.go
```diff
@@ -37,21 +37,23 @@ func main() {
...
 	switch command {
 	case "encrypt":
-		encrypt.Encrypt(args)
+		err = encrypt.Encrypt(args)
 	case "decrypt":
-		decrypt.Decrypt(args)
+		err = decrypt.Decrypt(args)
 	case "download":
-		download.Download(args)
+		err = download.Download(args)
```
jbygdell, line 48: Should this change be here? it is not related to encryption

main.go
```diff
@@ -37,21 +37,23 @@ func main() {
...
 
+	var err error = nil
+
 	switch command {
 	case "encrypt":
-		encrypt.Encrypt(args)
+		err = encrypt.Encrypt(args)
 	case "decrypt":
-		decrypt.Decrypt(args)
+		err = decrypt.Decrypt(args)
```
jbygdell, line 46: Should this change be here? it is not related to encryption

main.go
```diff
@@ -37,21 +37,23 @@ func main() {
...
 	case "download":
-		download.Download(args)
+		err = download.Download(args)
 	case "upload":
-		err := upload.Upload(args)
-		if err != nil {
-			log.Fatal(err)
-		}
+		err = upload.Upload(args)
```
jbygdell, line 50: Should this change be here? it is not related to encryption

jbygdell (changes requested):
Make a habit of using the tools provided by the GO installation before you make a commit.
gofmt
goimports
golint
golangci-lint

Then you can fix these easy to miss things:
```
encrypt/encrypt.go:21:5: exported var Usage should have comment or be unexported
encrypt/encrypt.go:33:5: exported var ArgHelp should have comment or be unexported
encrypt/encrypt.go:37:5: exported var Args should have comment or be unexported
encrypt/encrypt.go:43:1: comment on exported function Encrypt should be of the form "Encrypt ..."
encrypt/encrypt.go:302:1: comment on exported function FileExists should be of the form "FileExists ..."
encrypt/encrypt.go:310:1: comment on exported function FileIsReadable should be of the form "FileIsReadable ..."
```
and
```
encrypt/encrypt.go:133:38: Error return value of `ChecksumFileUnencSha256.WriteString` is not checked (errcheck)
                ChecksumFileUnencSha256.WriteString(
                                                   ^
encrypt/encrypt.go:135:33: Error return value of `ChecksumFileEncMd5.WriteString` is not checked (errcheck)
                ChecksumFileEncMd5.WriteString(
                                              ^
encrypt/encrypt.go:137:36: Error return value of `ChecksumFileEncSha256.WriteString` is not checked (errcheck)
                ChecksumFileEncSha256.WriteString(
```

encrypt/encrypt.go
norling: in this context, it's the private key that's given to the function, so it's correct. :) But I'll see if I can be more clear in the future

encrypt/encrypt.go
norling: Yes! definitely!

encrypt/encrypt.go
norling: Yes - but I wanna use it in the decrypt module as well, so I'm exporting it for that!

encrypt/encrypt.go
norling: same here - it's exported since I _do_ want to use it elsewhere.

decrypt/decrypt.go
norling: it's not. but it's a few lines to make all the modules behave the same, so I figured I could just include it. would you prefer if I break it out into a separate PR?

encrypt/encrypt.go
jbygdell: Isn't it better to put shared functions directly in main?

encrypt/encrypt.go
```diff
@@ -1,43 +1,338 @@
...
+	// encrypt the input files
+	numFiles := len(files)
+	for i, file := range files {
+		log.Infof("Encrypting file %v/%v: %s", i+1, numFiles, file.unencrypted)
+
+		// encrypt the file
+		err = encrypt(file.unencrypted, file.encrypted, *publicKey, *privateKey)
+		if err != nil {
+			return err
+		}
```
jbygdell, line 129:
```suggestion
		if err = encrypt(file.unencrypted, file.encrypted, *publicKey, *privateKey); err != nil {
			return err
		}
```

Just to be consistent with the same type of function call

encrypt/encrypt.go
```diff
@@ -1,43 +1,338 @@
...
+
+		files = append(files, encryptionFileSet{filename, outFilename})
+	}
+
+	// Check that all the infiles exist, and all the outfiles don't
+	err = checkFiles(files)
+	if err != nil {
+		return err
+	}
```
jbygdell, line 80:
```suggestion
	if err = checkFiles(files); err != nil {
		return err
	}
```

jbygdell (changes requested):
You tend to break lines in function calls out of habit rather when it is required for readability.

encrypt/encrypt_test.go
```diff
@@ -0,0 +1,182 @@
...
+	// unencrypted is readable, but encrypted exists
+	test_has_encrypted := encryptionFileSet{suite.file_ok.Name(),
+		suite.file_ok.Name()}
+	err = checkFiles([]encryptionFileSet{test_has_encrypted})
+	assert.EqualError(suite.T(), err, fmt.Sprintf("outfile %s already exists",
+		suite.file_ok.Name()))
+
+	// unencrypted isn't readable
+	test_no_unencrypted := encryptionFileSet{"does-not-exist",
+		suite.file_ok.Name()}
```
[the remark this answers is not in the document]
norling: My code windows are 80 characters so for me it's super annoying that I can't break lines... but I'll do it if needed ^^

dbampalikis (approved):
Looks good!

jbygdell (approved):

encrypt/encrypt.go
```diff
@@ -1,43 +1,350 @@
...
+	log.Info("Reading Public key file")
+	file, err := os.Open(filepath.Clean(filename))
+	if err != nil {
+		return nil, err
+	}
+
+	// This function panics if the key is malformed, so we handle that as well
+	// as errors
+	defer func() {
+		log.Info("Hey! this is a panic!")
```
aaperis, line 261:
```suggestion
```
This message prints anyways. Either remove or move it inside the if statement.

aaperis (approved):
Looks very good!
consider relieving unnecessary panicking :-)

encrypt/encrypt.go
norling: yes :D I just found it myself! I'll remove it ^^

PR #3, opened 2022-01-24
Add upload module

dbampalikis:
Add upload module to the tool.
- Read configuration from auth endpoint files
- Upload list of files
- Unit tests
Integration tests are coming in the next pull request


go.mod
```diff
@@ -4,4 +4,12 @@ go 1.17
...
-require golang.org/x/sys v0.0.0-20191026070338-33540a1f6037 // indirect
+require (
+	github.com/aws/aws-sdk-go v1.42.38
+	golang.org/x/sys v0.0.0-20210423082822-04245dca01da // indirect
+)
+
+require (
+	github.com/jmespath/go-jmespath v0.4.0 // indirect
+	gopkg.in/ini.v1 v1.66.3
+)
```
jbygdell, line 15: merge these two

upload/upload.go
```diff
@@ -21,7 +27,42 @@ var ArgHelp = `
...
+// Configuration struct for storing the s3cmd file values
+type Config struct {
+	AccessKey            string `ini:"access_key"`
+	SecretKey            string `ini:"secret_key"`
+	AccessToken          string `ini:"access_token"`
+	CheckSslCertificate  bool   `ini:"check_ssl_certificate"`
+	CheckSslHostname     bool   `ini:"check_ssl_hostname"`
+	SocketTimeout        int    `ini:"socket_timeout"`
+	Encoding             string `ini:"encoding"`
+	Encrypt              bool   `ini:"encrypt"`
```
jbygdell, line 41: this is somewhat redundant since the data is encrypted

upload/upload.go
```diff
@@ -21,7 +27,42 @@ var ArgHelp = `
...
+	HumanReadableSizes   bool   `ini:"human_readable_sizes"`
+	MultipartChunkSizeMb int64  `ini:"multipart_chunk_size_mb"`
+	HostBucket           string `ini:"host_bucket"`
+	HostBase             string `ini:"host_base"`
+	GuessMimeType        bool   `ini:"guess_mime_type"`
+	UseHttps             bool   `ini:"use_https"`
```
jbygdell, line 47: most of these should have sane defaults

dbampalikis: I thought that it would be nice to keep all the configuration there, in case we want to change something in the future. I could though read in only the ones we are actually using for the uploading. What would be the best?

jbygdell: Encrypt can be removed as that requires that the S3 backend has a crypto module.
give the rest production ready default values that is not backend specific. i.e. igore Hostbase and HostBucket

jbygdell: MultipartChunkSizeMb is default 15Mb, but for us it needs to be bigger, 50-100Mb

jbygdell: the booleans should default to `true`

dbampalikis: I think it makes sense to leave only the values that are needed: credentials, Hostbase/bucket and chunksize. And I will remove all the rest

upload/upload.go
```diff
@@ -54,41 +53,63 @@ func loadConfigFile(path string) (*Config, error) {
...
 	if err != nil {
-		return config, nil
+		return config, err
 	}
 
 	if err := cfg.Section("").MapTo(config); err != nil {
 		return nil, err
 	}
 
+	if config.AccessKey == "" || config.SecretKey == "" || config.AccessToken == "" {
```
jbygdell, line 63: `SecretKey` is strictly not necessary since authentication is done with the token and `Accesskey` should be the username

upload/upload.go
```diff
@@ -54,41 +53,63 @@ func loadConfigFile(path string) (*Config, error) {
...
+		return nil, errors.New("failed to find credentials in configuration file")
+	}
+
+	if config.HostBase == "" {
+		config.HostBase = "inbox.test.ega.nbis.se"
```
jbygdell, line 68: This can not be hardcoded.

upload/upload.go
```diff
@@ -54,41 +53,63 @@ func loadConfigFile(path string) (*Config, error) {
...
+
+	if config.UseHttps {
+		config.HostBase = "https://" + config.HostBase
+	}
+
+	if config.Encoding == "" {
+		config.Encoding = "UTF-8"
+	}
+
+	if config.MultipartChunkSizeMb == 0 {
```
jbygdell, line 79:
```suggestion
	if config.MultipartChunkSizeMb <= 15 {
```

upload/upload_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	defer os.Remove(configPath.Name())
+
+	err = ioutil.WriteFile(configPath.Name(), []byte(confFile), 0644)
+	if err != nil {
+		log.Fatalf("failed to write temp config file, %v", err)
+	}
+
+	_, err = loadConfigFile(configPath.Name())
```
jbygdell, line 94:
```suggestion
	configPath, err := ioutil.TempFile(os.TempDir(), "s3cmd-")
	if err != nil {
		log.Fatal(err)
	}

	_, err = loadConfigFile(configPath.Name())
```

jbygdell: No need for the write delete dance, just create an empty file.

upload/upload_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (suite *TestSuite) TestConfig() {
+	var confFile = `
+access_token = eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2NDM0ODQxNTIsImlzcyI6Imh0dHBzOi8vbG9naW4udGVzdC5lZ2EubmJpcy5zZSIsInN1YiI6ImVnYS1ib3gtODkwMiJ9.JS-R4GUODAzbYzYs1hJR_92D9I3blfcGu5_yrK75L6QnIK6Aq6Po6kBbngJa-N9SRIzETxy5xd5uZthHaOQpFQ
+host_base = inbox.test.ega.nbis.se
+encoding = UTF-8
+host_bucket = inbox.test.ega.nbis.se
+multipart_chunk_size_mb = 50
+secret_key = ega-box-8902
+access_key = ega-box-8902
```
jbygdell, line 106: Don't hard code hosts or credentials to the staging system

upload/upload_test.go
```diff
@@ -0,0 +1,173 @@
...
+		log.Fatal(err)
+	}
+
+	defer os.Remove(configPath.Name())
+	log.Print(configPath.Name())
+
+	err = ioutil.WriteFile(configPath.Name(), []byte(confFile), 0644)
+	if err != nil {
+		log.Fatalf("failed to write temp config file, %v", err)
+	}
```
jbygdell, line 167: Since the config file is used by multiple instances put it in a struct so there are only one place to change stuff, then each test can change things to be tested before it writes it to a temp file at the beginning of the test.

jbygdell (changes requested):

upload/upload.go
dbampalikis: Should I create a configuration then including the `HostBase` and `region`?

upload/upload_test.go
dbampalikis: oh yes you are right on here, no need to actually write something in the file

upload/upload.go
jbygdell: Standalone SDA, federated EGA as well as all test sites need to be able to use this tool.
So anything that corresponds to host names, user ID and so on can not be but in the code.

upload/upload_test.go
dbampalikis: If we don't want to have the host in the file, I guess I will have to move the not authorized test to the integration ones and use the s3-Proxy with minio

jbygdell: This test should not have to talk to anything outside your computer.

jbygdell: If you need anything from S3 you have to mock it.

upload/upload.go
dbampalikis: Why less than 15 here? With 0 I make sure I give a value, if it's not set in the config file

upload/upload.go
```diff
@@ -21,18 +27,119 @@ var ArgHelp = `
...
+		// Upload the file to S3.
+		result, err := uploader.Upload(&s3manager.UploadInput{
+			Body:            f,
+			Bucket:          aws.String(config.AccessKey),
+			Key:             aws.String(filename),
+			ContentEncoding: aws.String(config.Encoding),
+		}, func(u *s3manager.Uploader) {
+			u.PartSize = config.MultipartChunkSizeMb * 1024 * 1024
+			// Delete parts of failed multipart, since we cannot currently continue them
+			u.LeavePartsOnError = true
```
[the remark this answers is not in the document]
dbampalikis: I think that we have an issue with continuing the multipart currently, so I left a comment and we can change it once we have it fixed in the s3 proxy

upload/upload.go
```diff
@@ -21,18 +27,119 @@ var ArgHelp = `
...
+	if err := cfg.Section("").MapTo(config); err != nil {
+		return nil, err
+	}
+
+	if config.AccessKey == "" || config.SecretKey == "" || config.AccessToken == "" {
+		return nil, errors.New("failed to find credentials in configuration file")
+	}
+
+	if config.HostBase == "" {
+		config.HostBase = "inbox.test.ega.nbis.se"
```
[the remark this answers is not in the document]
jbygdell: We shouldn't default to anything, unless we decide to create one binary fore each archive, EGA, SDA, test and so on.

upload/upload.go
jbygdell: Library default is 15Mb and max filesize is 10000 x MultipartChunkZise

upload/upload_test.go
```diff
@@ -0,0 +1,131 @@
...
+	defer os.Remove(configPath.Name())
+
+	err = ioutil.WriteFile(configPath.Name(), []byte(confFile), 0644)
+	if err != nil {
+		log.Fatalf("failed to write temp config file, %v", err)
+	}
+
+	_, err = loadConfigFile(configPath.Name())
+	assert.EqualError(suite.T(), err, "failed to find endpoint in configuration file")
+}
```
jbygdell, line 90: maybe we should also test missing value i.e. 
```
secret_key = 
```

upload/upload.go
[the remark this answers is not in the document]
jbygdell: if `LeavePartsOnError = true` the failed upload will *not* be cleaned up automatically.

upload/upload.go
jbygdell: If multipart size is not in the config file it will be set to 15Mb by the library, so we only needs to guard ourselves from the case when the config file has a value less than 15

jbygdell (approved):
looking much better, some comments only

upload/upload_test.go
dbampalikis: That's a good point. Maybe I should have a readme with the expected file instead. What do you think?

PR #4, opened 2022-01-26
Gihub actions to run tests

jbygdell:


norling (approved):
I had planned to have the modules return their errors to be handled by main, but I see you added a `log.Fatal` to the argument parsing. Do you prefer to have the modules not return errors? Either is fine with me :)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,44 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
```
aaperis, line 5:
```suggestion
    branches: [main]
```
not sure if this essential but still..

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,44 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
```
aaperis, line 8:
```suggestion
    branches: [main]
```
same as above

aaperis (approved):
looks good :-)

PR #8, opened 2022-01-26
Feature/download

dbampalikis:
Add download module
- Gets the `urls_list.txt` file or a url to the containing folder in our s3Backend (ending with `/`)
- Downloads the files from the s3 bucket creating the same folder structure

Tests are following


jbygdell (changes requested):
It really is no need for log.fatal in a test.
Other than that it looks fine.

download/download_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	assert.EqualError(suite.T(), err, "failed to download file, reason: Get \"someUrl\": unsupported protocol scheme \"\"")
+}
+
+func (suite *TestSuite) TestWrongUrlsFile() {
+
+	urlsListPath, err := ioutil.TempFile(os.TempDir(), "urls_list-")
+	if err != nil {
+		log.Fatal(err)
+	}
```
jbygdell, line 47:
```suggestion
	assert.NoError(suite.T(), err)
```

the log Fatal will kill the process making the defer on the next lint to not run.

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+	// The folder structure comes after the UID described in the regex
+	re := regexp.MustCompile(`(?i)[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/(.*)`)
+	match := re.FindStringSubmatch(file)
+	if match == nil || len(match) < 1 {
+		return fileName, fmt.Errorf("failed to parse url for downloading file")
+	}
+	fileName = match[1]
+
+	var filePath string
+	if strings.Contains(fileName, "/") {
```
norling, line 52: I would probably move this folder creation stuff outside the function, as it's not obvious that this happens from the function name.

norling (approved):
I had a comment about having folder creation code in `getFileNameFromURL()`, but looks good otherwise! :)

download/download.go
dbampalikis: Since they are needed for the folder creation, I thought it makes sense to take them away from the main function. So the function now is getting the url, exports the file and creates the folder where the file will be stored. But if the consensus is to move it, I am fine with it.

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
         all flagless arguments will be used as download urls.`
 
+// Args is a flagset that needs to be exported so that it can be written to the
+// main program help
 var Args = flag.NewFlagSet("download", flag.ExitOnError)
 
-// Main download function
-func Download(args []string) {
+// Gets the file name for a URL, using regex
+func getFileNameFromURL(file string) (fileName string, err error) {
```
jbygdell, line 41:
```suggestion
func createFilePathFromURL(file string) (fileName string, err error) {
```
Changing the name to better match what is going on.

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+	// Open urls_list.txt file and loop through file urls
+	urlsList, err := getFilesUrls(urlsFilePath)
+	if err != nil {
+		return err
+	}
+
+	// Download the files and create the folder structure
+	for _, file := range urlsList {
+
+		fileName, err := getFileNameFromURL(file)
```
jbygdell, line 165:
```suggestion
		fileName, err := createFilePathFromURL(file)
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+		if err != nil {
+			return fileName, err
+		}
+	}
+
+	return fileName, nil
+}
+
+// Downloads a file from the url to the filePath location
+func downloadListFile(url string, filePath string) error {
```
jbygdell, line 64:
```suggestion
func downloadFile(url string, filePath string) error {
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+
+	// Download the files and create the folder structure
+	for _, file := range urlsList {
+
+		fileName, err := getFileNameFromURL(file)
+		if err != nil {
+			return err
+		}
+
+		err = downloadListFile(file, fileName)
```
jbygdell, line 170:
```suggestion
		err = downloadFile(file, fileName)
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+	// Write the body to file
+	_, err = io.Copy(out, resp.Body)
+	defer out.Close()
+
+	return err
+
+}
+
+// Read the urls_list.txt file and return the urls of the files in a list
+func getFilesUrls(urlsFilePath string) (urlsList []string, err error) {
```
jbygdell, line 89:
```suggestion
func getURLsFile(urlsFilePath string) (urlsList []string, err error) {
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+			return err
+		}
+	// Case where the user passes a file containg the urls to download
+	// e.g. /some/folder/to/file.txt
+	default:
+		urlsFilePath = urls[0]
+	}
+
+	// Open urls_list.txt file and loop through file urls
+	urlsList, err := getFilesUrls(urlsFilePath)
```
jbygdell, line 157:
```suggestion
	urlsList, err := getURLsFile(urlsFilePath)
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+		log.Info(urls[0], urlsFilePath)
+		err = downloadListFile(urls[0], urlsFilePath)
+		if err != nil {
+			return err
+		}
+	// Case where the user passes a file containg the urls to download
+	// e.g. /some/folder/to/file.txt
+	default:
+		urlsFilePath = urls[0]
+	}
```
jbygdell, line 154: No case where the user can supply the URL to a single datafile ?
we can not assume that the user will always download all files every time.

jbygdell (changes requested):
Code looks fine, function names on the other hand could have been better at reflecting what's actually being done.

download/download.go
dbampalikis: That's covered but leaving only the wanted files in the `urls_list.txt` file. Shouldn't that be enough?

download/download_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	_, err = getURLsFile(urlsListPath.Name())
+	assert.EqualError(suite.T(), err, "failed to get list of files, empty file")
+}
+
+func (suite *TestSuite) TestCorrectUrlsFile() {
+
+	urlsListFile := `someUrlToFile1
+someUrlToFile2
+someUrlToFile3
```
aaperis, line 56:
```suggestion
someUrlToFile3
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+
+	if len(urlsList) == 0 {
+		return urlsList, fmt.Errorf("failed to get list of files, empty file")
+	}
+
+	return urlsList, scanner.Err()
+}
+
+// Download function downaloads the files included in the urls_list.txt file.
```
aaperis, line 109:
```suggestion
// Download function downloads the files included in the urls_list.txt file.
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
...
+	}
+
+	var currentPath, urlsFilePath string
+	currentPath, err = os.Getwd()
+	if err != nil {
+		return fmt.Errorf("failed to get current path, reason: %v", err)
+	}
+
+	switch {
+	// Case where the uses passes the url to the s3 folder where the data exists
```
aaperis, line 132:
```suggestion
	// Case where the user passes the url to the s3 folder where the data exists
```

download/download.go
```diff
@@ -17,23 +26,155 @@ Download: Downloads files from the Sensitive Data Archive (SDA). If a directory
           download the urls_list.txt file, and then download all files in this
           list. If file urls are given, they will be downloaded as-is.
```
aaperis, line 27: Last statement is somewhat confusing from a user perspective. Maybe give more detail for the other cases covered in the code instead?

aaperis (approved):
looks good :-), minor comments

kostas-kou (approved):
Looks good to me

shreyasshivakumara (approved):

PR #12, opened 2022-01-28
Feature/decryption

norling:
This will likely need to be rebased once download is merged, so I'll keep it as a draft in the mean time.

### DIFF
--- a/decrypt/decrypt.go
+++ b/decrypt/decrypt.go
@@ -147,7 +147,6 @@ func readPrivateKey(filename, password string) (key *[32]byte, err error) {
 	// This function panics if the key is malformed, so we handle that as well
 	// as errors
 	defer func() {
-		log.Info("Hey! this is a panic!")
 		if recover() != nil {
 			err = fmt.Errorf("malformed key file: %s", filename)
 		}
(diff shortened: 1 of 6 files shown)


decrypt/decrypt.go
```diff
@@ -1,43 +1,248 @@
...
+	defer func() {
+		if err := secFile.Close(); err != nil {
+			log.Errorf("Error closing file: %s\n", err)
+		}
+	}()
+	pass := []byte(password)
+	err = keys.WriteCrypt4GHX25519PrivateKey(secFile, privateKeyData, pass)
+
+	return err
+}
```
jbygdell, line 158:
```suggestion
```

this is probably a copy paste error, since by now you should already have your keypair.

decrypt/decrypt.go
```diff
@@ -1,43 +1,248 @@
...
 	log "github.com/sirupsen/logrus"
 )
 
 // Help text and command line flags.
 
+// Usage text that will be displayed as command line help text when using the
+// `help decrypt` command
 var Usage = `
-USAGE: %s decrypt -key <private-key-file> [file(s)]
+USAGE: %s decrypt (-createKey <name>) -key <private-key-file> [file(s)]
```
jbygdell, line 23:
```suggestion
USAGE: %s decrypt  -key <private-key-file> [file(s)]
```

this module doesn't have any need to create a keypair

decrypt/decrypt.go
norling: yes :D

PR #17, opened 2022-01-31
Download path

dbampalikis:
Added outDir flag to download module
- Allow the user to select the path where the data should be downloaded

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -27,7 +27,6 @@ Download: Downloads files from the Sensitive Data Archive (SDA). The files will
 		  to first download the urls_list.txt file, and then download all files in 
 		  this list. If file urls are given, they files will be downloaded creating 
 		  the same folder structure locally.
->>>>>>> Added outDir flag to download module
 `
 
 // ArgHelp is the suffix text that will be displayed after the argument list in
--- a/download/download.go
+++ b/download/download.go
@@ -22,11 +22,11 @@ var Usage = `
 USAGE: %s download (-outdir <dir>) [url(s) | file]
 
 Download: Downloads files from the Sensitive Data Archive (SDA). The files will
-		  be downloaded in the current directory, if outdir is not defined.
-		  If a directory is provided (ending with "/"), then the tool will attempt 
-		  to first download the urls_list.txt file, and then download all files in 
-		  this list. If file urls are given, they files will be downloaded creating 
-		  the same folder structure locally.
+	  be downloaded in the current directory, if outdir is not defined.
+	  If a directory is provided (ending with "/"), then the tool will attempt 
+	  to first download the urls_list.txt file, and then download all files in 
+	  this list. If file urls are given, they files will be downloaded creating 
+	  the same folder structure locally.
 `
 
 // ArgHelp is the suffix text that will be displayed after the argument list in
--- a/download/download.go
+++ b/download/download.go
@@ -19,13 +19,15 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help download` command
 var Usage = `
-USAGE: %s download [url(s)]
+USAGE: %s download (-outdir <dir>) [url(s) | file]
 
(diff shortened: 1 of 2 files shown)


download/download.go
```diff
@@ -19,13 +19,14 @@ import (
...
-          list. If file urls are given, they files will be downloaded creating the
-		  same folder structure locally.
+USAGE: %s download (-outdir <dir>) [url(s) | file]
+
+Download: Downloads files from the Sensitive Data Archive (SDA). The files will
+          be downloaded in the current directory, if outdir is not defined.
+		  If a directory is provided (ending with "/"), then the tool will attempt 
+		  to first download the urls_list.txt file, and then download all files in 
+		  this list. If file urls are given, they files will be downloaded creating 
+		  the same folder structure locally.
```
norling, line 29: There seems to be tabs instead of spaces here, messing up the indentation
```suggestion
          If a directory is provided (ending with "/"), then the tool will attempt 
          to first download the urls_list.txt file, and then download all files in 
          this list. If file urls are given, they files will be downloaded creating 
          the same folder structure locally.
```

download/download.go
```diff
@@ -37,17 +38,21 @@ var ArgHelp = `
...
 	re := regexp.MustCompile(`(?i)[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/(.*)`)
 	match := re.FindStringSubmatch(file)
 	if match == nil || len(match) < 1 {
 		return fileName, fmt.Errorf("failed to parse url for downloading file")
 	}
-	fileName = match[1]
+	if baseDir != "" && !strings.HasSuffix(baseDir, "/") {
+		baseDir += "/"
+	}
+	fileName = baseDir + match[1]
```
norling, line 55: It's not important, but might be neater to do this like:
```
    fileName  = filepath.Join(baseDir, match[1])
```
That's probably more platform safe as well.

download/download.go
```diff
@@ -154,6 +159,7 @@ func Download(args []string) error {
 		urlsFilePath = urls[0]
 	}
 
+	log.Print(urlsFilePath)
```
norling, line 162: Just checking that this log is supposed to be here, and not a forgotten debug log? ^^

norling (approved):
Some comments, but nothing major!

jbygdell (changes requested):

download/download.go
```diff
@@ -19,13 +19,14 @@ import (
...
-          list. If file urls are given, they files will be downloaded creating the
-		  same folder structure locally.
+USAGE: %s download (-outdir <dir>) [url(s) | file]
+
+Download: Downloads files from the Sensitive Data Archive (SDA). The files will
+		  be downloaded in the current directory, if outdir is not defined.
+		  If a directory is provided (ending with "/"), then the tool will attempt 
+		  to first download the urls_list.txt file, and then download all files in 
+		  this list. If file urls are given, they files will be downloaded creating 
+		  the same folder structure locally.
```
jbygdell, line 29:
```suggestion
Download: Downloads files from the Sensitive Data Archive (SDA). The files will
          be downloaded in the current directory, if outdir is not defined.
          If a directory is provided (ending with "/"), then the tool will attempt 
          to first download the urls_list.txt file, and then download all files in 
          this list. If file urls are given, they files will be downloaded creating 
          the same folder structure locally.
```

This is annoying, tabspace in a terminal is the equivalent of 8 spaces.
That means that the view in GH will never be aligned if it looks good in the console output. 😞

dbampalikis: Yes I just noticed that too. In VS it looks fine, but when checking it in the terminal it looks weird :/

jbygdell: You can set the tabspace to 8 spaces in VS code

jbygdell: Then it look fine in  both

jbygdell (approved):

PR #19, opened 2022-02-01
add createKey module

norling:
Adds a (now separate) create-key module.

### DIFF
--- a/create_key/create_key.go
+++ b/create_key/create_key.go
@@ -61,8 +61,14 @@ func CreateKey(args []string) error {
 	// Add the output directory to the file path (does nothing if outDir is "")
 	basename = filepath.Join(*outDir, basename)
 
+	// Read password from user, to avoid having it in plaintext as an argument
+	password, err := helpers.PromptPassword("Enter private key password")
+	if err != nil {
+		return fmt.Errorf("failed to read password from user: %v", err)
+	}
+
 	// Write the key files
-	err = generateKeyPair(basename, "")
+	err = generateKeyPair(basename, password)
 
 	return err
 }
--- a/create_key/create_key.go
+++ b/create_key/create_key.go
@@ -75,11 +75,8 @@ func generateKeyPair(basename, password string) error {
 	publicKeyName := fmt.Sprintf("%s.pub.pem", basename)
 
 	// check if any of the files exist
-	if helpers.FileExists(publicKeyName) {
-		return fmt.Errorf("public key file %s already exists, refusing to overwrite", publicKeyName)
-	}
-	if helpers.FileExists(privateKeyName) {
-		return fmt.Errorf("private key file %s already exists, refusing to overwrite", privateKeyName)
+	if helpers.FileExists(publicKeyName) || helpers.FileExists(privateKeyName){
+		return fmt.Errorf("Key pair with name '%v' seems to already exist, refusing to overwrite", basename)
 	}
 
 	// Generate key pair
--- a/create_key/create_key.go
+++ b/create_key/create_key.go
@@ -75,7 +75,7 @@ func generateKeyPair(basename, password string) error {
 	publicKeyName := fmt.Sprintf("%s.pub.pem", basename)
 
 	// check if any of the files exist
-	if helpers.FileExists(publicKeyName) || helpers.FileExists(privateKeyName){
+	if helpers.FileExists(publicKeyName) || helpers.FileExists(privateKeyName) {
 		return fmt.Errorf("Key pair with name '%v' seems to already exist, refusing to overwrite", basename)
 	}
 
--- a/create_key/create_key.go
+++ b/create_key/create_key.go
@@ -111,8 +111,7 @@ func generateKeyPair(basename, password string) error {
 			log.Errorf("Error closing file: %s\n", err)
 		}
 	}()
-	pass := []byte(password)
-	err = keys.WriteCrypt4GHX25519PrivateKey(secFile, privateKeyData, pass)
+	err = keys.WriteCrypt4GHX25519PrivateKey(secFile, privateKeyData, []byte(password))
 
 	return err
 }
--- a/create_key/create_key.go
+++ b/create_key/create_key.go
@@ -58,39 +58,13 @@ func CreateKey(args []string) error {
 	}
 	basename := Args.Args()[0]
 
-	// If the `outdir` flag is set, change to the output directory before
-	// writing the key files.
-	startDir := ""
-	if *outDir != "" {
-		startDir, err = os.Getwd()
-		if err != nil {
-			return fmt.Errorf("could not get current directory: %v", err)
-		}
-		err = os.Chdir(*outDir)
-		if err != nil {
-			return fmt.Errorf("could not enter output directory: %v", err)
-		}
-	}
+	// Add the output directory to the file path (does nothing if outDir is "")
+	basename = filepath.Join(*outDir, basename)
 
 	// Write the key files
 	err = generateKeyPair(basename, "")
-	if err != nil {
-		return err
-	}
 
-	// If we changed to a different directory for writing files, we change back
-	// in case some other function is called later which assumes the working
-	// directory hasn't changed.
-	if *outDir != "" {
-		err = os.Chdir(startDir)
-		if err != nil {
-			return fmt.Errorf("could not return to start directory: %v", err)
-		}
-
-		return err
-	}
-
-	return nil
+	return err
 }
 
 // Generates a crypt4gh key pair, and saves it to the `<basename>.pub.pem` and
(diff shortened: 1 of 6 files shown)


create_key/create_key.go
```diff
@@ -0,0 +1,147 @@
...
+// `help encrypt` command
+var Usage = `
+USAGE: %s createKey <name> (-outdir <dirname>)
+
+createKey: Creates a crypt4gh encryption key pair, and saves it to
+           <name>.pub.pem, and <name>.sec.pem.
+           NOTE: keys created using this function should not be used when
+		         encrypting submission files, they should only be used for
+                 decrypting files downloaded from the archive.
+`
```
jbygdell, line 28:
```suggestion
createKey: Creates a crypt4gh encryption key pair, and saves it to
		  <name>.pub.pem, and <name>.sec.pem.
		  NOTE: keys created using this function should not be used when
		  encrypting submission files, they should only be used for
		  decrypting files downloaded from the archive.
`
```

create_key/create_key.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Args is a flagset that needs to be exported so that it can be written to the
+// main program help
+var Args = flag.NewFlagSet("createKey", flag.ExitOnError)
+
+var outDir = Args.String("outdir", "", "Output directory for the key files")
+
+// Encrypt takes a set of arguments, parses them, and attempts to encrypt the
+// given data files with the given public key file
+func CreateKey(args []string) error {
```
jbygdell, line 42: Sloppy copy-paste error

create_key/create_key.go
```diff
@@ -0,0 +1,147 @@
...
+// Usage text that will be displayed as command line help text when using the
+// `help encrypt` command
+var Usage = `
+USAGE: %s createKey <name> (-outdir <dirname>)
+
+createKey: Creates a crypt4gh encryption key pair, and saves it to
+           <name>.pub.pem, and <name>.sec.pem.
+           NOTE: keys created using this function should not be used when
+		         encrypting submission files, they should only be used for
+                 decrypting files downloaded from the archive.
```
dbampalikis, line 27: I think there is something off with the indentation here, second line of NOTE seems to have one additional space

create_key/create_key.go
dbampalikis, line 42: Probably a copy pasting issue in the comment?

create_key/create_key.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	// If we changed to a different directory for writing files, we change back
+	// in case some other function is called later which assumes the working
+	// directory hasn't changed.
+	if *outDir != "" {
+		err = os.Chdir(startDir)
+		if err != nil {
+			return fmt.Errorf("could not return to start directory: %v", err)
+		}
```
dbampalikis, line 88: Instead of changing the working directory, I would pass the outdir to the `generateKeyPair` and create the key at that location. It's more a matter of preference I guess, but since we use it in download too, maybe it would be more consistent

create_key/create_key.go
```diff
@@ -0,0 +1,147 @@
...
+		return err
+	}
+	defer func() {
+		if err := secFile.Close(); err != nil {
+			log.Errorf("Error closing file: %s\n", err)
+		}
+	}()
+	pass := []byte(password)
+	err = keys.WriteCrypt4GHX25519PrivateKey(secFile, privateKeyData, pass)
+
```
jbygdell, line 145: Why don't you use the key generation function in the C4GH library?
https://github.com/elixir-oslo/crypt4gh/blob/b28572b08f0a3dd0a9cebb905ef2a563f8c3a0fa/main.go#L268

jbygdell (changes requested):

create_key/create_key.go
```diff
@@ -0,0 +1,147 @@
...
+	privateKeyName := fmt.Sprintf("%s.sec.pem", basename)
+	publicKeyName := fmt.Sprintf("%s.pub.pem", basename)
+
+	// check if any of the files exist
+	if helpers.FileExists(publicKeyName) {
+		return fmt.Errorf("public key file %s already exists", publicKeyName)
+	}
+	if helpers.FileExists(privateKeyName) {
+		return fmt.Errorf("private key file %s already exists", privateKeyName)
+	}
```
dbampalikis, line 109: I would check if both keys exist and then return a message that the user can now encrypt. With the current implementation, the user might assume that the key exists, even if only one of them is there

dbampalikis (changes requested):

create_key/create_key.go
norling: they're not exported :/

jbygdell: It is if you target the proper submodule
https://github.com/elixir-oslo/crypt4gh/blob/b28572b08f0a3dd0a9cebb905ef2a563f8c3a0fa/keys/keys.go#L44

jbygdell: Or you can just duplicate the code in the main.go from the library.

norling: I could - but I honestly don't want to. :) I don't want this script to be interactive as it's likely to be used by scripts, and getting the function options from flag is just hard to read imho.

jbygdell: Maybe not the code it self but the concept, currently you have duplicated every action.

dbampalikis (approved):

create_key/create_key_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	testFileName := filepath.Join(suite.tempDir, "keyfile")
+
+	// none of the target files exist
+	err := generateKeyPair(testFileName, "")
+	assert.NoError(suite.T(), err)
+
+	// now the targets exist - should crash on public-key existing
+	err = generateKeyPair(testFileName, "")
+	assert.EqualError(suite.T(), err, fmt.Sprintf("public key file %s.pub.pem already exists, refusing to overwrite", testFileName))
```
jbygdell, line 49:
```suggestion
	assert.EqualError(suite.T(), err, fmt.Sprintf("Key pair with name '%v' seems to already exist, refusing to overwrite", testFileName))
```

create_key/create_key_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// remove the public key to test the private key exists error
+	os.Remove(fmt.Sprintf("%s.pub.pem", testFileName))
+	err = generateKeyPair(testFileName, "")
+	assert.EqualError(suite.T(), err, fmt.Sprintf("private key file %s.sec.pem already exists, refusing to overwrite", testFileName))
```
jbygdell, line 54:
```suggestion
	assert.EqualError(suite.T(), err, fmt.Sprintf("Key pair with name '%v' seems to already exist, refusing to overwrite", testFileName))
```

create_key/create_key_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (suite *CreateKeyTests) SetupTest() {
+
+	var err error
+
+	// Create a temporary directory for our files
+	suite.tempDir, err = ioutil.TempDir(os.TempDir(), "sda-cli-test-")
+	if err != nil {
+		log.Fatal("Couldn't create temporary test directory", err)
+	}
```
jbygdell, line 32:
```suggestion
	assert.NoError(suite.T(), err)
```

create_key/create_key.go
```diff
@@ -0,0 +1,121 @@
...
+	if len(Args.Args()) < 1 {
+		return errors.New("no filename given")
+	}
+	basename := Args.Args()[0]
+
+	// Add the output directory to the file path (does nothing if outDir is "")
+	basename = filepath.Join(*outDir, basename)
+
+	// Write the key files
+	err = generateKeyPair(basename, "")
```
jbygdell, line 65: No password for the secret key?
How do you really think this tool is going to be used? entirely in a scripted environment?

jbygdell (changes requested):
Probably can't get any better than this

create_key/create_key.go
```diff
@@ -0,0 +1,121 @@
...
+	privateKeyName := fmt.Sprintf("%s.sec.pem", basename)
+	publicKeyName := fmt.Sprintf("%s.pub.pem", basename)
+
+	// check if any of the files exist
+	if helpers.FileExists(publicKeyName) {
+		return fmt.Errorf("public key file %s already exists, refusing to overwrite", publicKeyName)
+	}
+	if helpers.FileExists(privateKeyName) {
+		return fmt.Errorf("private key file %s already exists, refusing to overwrite", privateKeyName)
+	}
```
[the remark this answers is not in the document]
norling: sure :) I thought it was helpful to know which file was missing in case it was only one, but I get that's a pretty unlikely case ^^

create_key/create_key.go
norling: well - my goal was to make it as simple as possible, and since a lot of people won't give a meaningful password anyway, I figured we could skip it :) do you think we should enforce a password instead?

jbygdell: This app is what we say our users should use, then we should actively promote high security standards.
And since key creation is not something that will be scripted in most cases I say we enforce the use of passphrase protection.

norling: great! :) I'll make it passwordy!

jbygdell (approved):
Maybe add a test to verify that the generated key can be unlocked with the password used?
Other than that it looks fine

PR #27, opened 2022-02-08
Filesize module

dbampalikis:
The pull request add the file size functionality
- The user can check the sizes of the files, given a `urls_list.txt` file or the URL where the file is
- Added tests for file size


filesize/filesize.go
```diff
@@ -0,0 +1,103 @@
...
+	urlsFilePath, err = download.GetURLsListFile(currentPath, urls[0])
+	if err != nil {
+		return fmt.Errorf("failed to urls list file, reason: %v", err)
+	}
+
+	// Open urls_list.txt file and loop through file urls
+	urlsList, err := download.GetURLsFile(urlsFilePath)
+	if err != nil {
+		return err
+	}
```
jbygdell, line 87: Try not have duplication in the different parts of the app.
This is handled by another argument in the application.

filesize/filesize.go
```diff
@@ -0,0 +1,103 @@
...
+// Usage text that will be displayed as command line help text when using the
+// `help list` command
+var Usage = `
+USAGE: %s filesize [url(s) | file]
+
+Filesize: List files that can be downloaded from the Sensitive Data Archive (SDA).
+	  If a directory is provided (ending with "/"), then the tool will attempt 
+	  to first download the urls_list.txt file, and then return a list of the
+	  files with their respective sizes.
+`
```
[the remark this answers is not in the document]
dbampalikis: Yeah that's a good point. Do you think I should change only the name of the command or the package and the file names?

filesize/filesize.go
dbampalikis: I am not sure I follow this comment. Could you be a bit more specific?

jbygdell: What I'm talking about is the fact that lot's of unintended stuff is happening by magic, in this case it is downloading a file when the expected action is to display the size of a file that can be downloaded.

filesize/filesize.go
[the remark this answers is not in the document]
jbygdell: If the idea is that is shall return the size of all files in the dataset, maybe datasetsize is a better name.
Returning the sizes of all files as well as the summed total.

filesize/filesize.go
jbygdell: While we need the text file in order to get the filenames for all the other files there us no need for it to be saved to disk in this setting.

dbampalikis: The idea with saving the file (as it happens in the download module) is that I could have the same code for all the cases: when the user has the file already in their disk and when they pass a url to the file or to the containing folder. If that's not the case, there should be different code paths for files that exist in the disk and others that are supposed to be read with a request

aaperis (approved):
looks good to me :-)

norling (approved):
I think it looks good! :)

PR #28, opened 2022-02-09
Token expiration

### LINKED ISSUE #24: Show token information on upload
When uploading a file (and potentially as a callable module) the system should be able to report if a token is valid, and it's remaining lifetime.

dbampalikis:
Add token expiration check in upload.
- Checks if the token expires in less than 24 hours and notify user
- Add tests for expiration function

Closes #24

### DIFF
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -105,11 +105,6 @@ func checkTokenExpiration(accessToken string) (bool, error) {
 		return false, fmt.Errorf("could not parse token, reason: %s", err)
 	}
 
-	// Return error if token is broken (without claims)
-	if claims, ok := token.Claims.(jwt.MapClaims); !ok {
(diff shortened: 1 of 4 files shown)


upload/upload.go
```diff
@@ -93,6 +96,41 @@ func loadConfigFile(path string) (*Config, error) {
...
+		switch iat := claims["exp"].(type) {
+		case float64:
+			expiration = time.Unix(int64(iat), 0)
+		case json.Number:
+			tmp, _ := iat.Int64()
+			expiration = time.Unix(tmp, 0)
+		}
+
+	}
+
```
jbygdell, line 128: These 2 could have been concatenated into an if-else clause since they check the same thing.

jbygdell (approved):

upload/upload.go
dbampalikis: You are right, fixed now

aaperis (approved):
looks good :-)

PR #38, opened 2022-03-08
Check if a file in the input list is already in c4gh format

shreyasshivakumara:


encrypt/encrypt.go
```diff
@@ -73,6 +73,11 @@ func Encrypt(args []string) error {
 			outFilename = path.Join(*outDir, basename) + ".c4gh"
 		}
 
+		// check that the encrypted file already exists
+		if helpers.FileExists(outFilename) {
+			return fmt.Errorf("Encrypted filename %s already exists", outFilename)
+		}
```
jbygdell, line 79:
```suggestion
		// check that the encrypted file already exists
		if helpers.FileExists(outFilename) {
			return fmt.Errorf("Encrypted filename %s already exists", outFilename)
		}
```

jbygdell (changes requested):

encrypt/encrypt.go
```diff
@@ -73,6 +73,11 @@ func Encrypt(args []string) error {
...
+
```
aaperis, line 80: This fix does not seem to resolve the issue. 
For example, the output of  `./sda-cli encrypt -key c4gh.pub.pem dummy_data.c4gh` is `dummy_data.c4gh.c4gh`. Instead of checking if the file exists, it would be preferable to check if the input file is `*.c4gh`. Maybe even better, if the first 8 bytes of the input file are `crypt4gh`.

aaperis (changes requested):

encrypt/encrypt.go
shreyasshivakumara: Right, I will change it.

dbampalikis: Also, you could add a test about it in the respective tests file

shreyasshivakumara: Yes! On it!

norling (changes requested):
I'd like it if you move your file checks into the `checkFiles` function to keep all of the checks in the same place. In there there's also a check for if the infile exists, and outfile doesn't, so you only need to add your extension check!

encrypt/encrypt.go
```diff
@@ -66,13 +66,29 @@ func Encrypt(args []string) error {
...
+		// Input filename is not found
+		if !helpers.FileExists(filename) {
+			return fmt.Errorf("Input filename is not found")
+		}
+
+		// Check if the input filename is already encrypted
+		fileExtension := filepath.Ext(filename)
+		if fileExtension == ".c4gh" {
+			return fmt.Errorf("Input filename is already encrypted(.c4gh)")
+		}
```
jbygdell, line 78: This doesn't verify that the file is encrypted, only that it has a `.c4gh` file ending, which by it self is not enough.
Crypt4gh files have a `magic number` at the beginning of the header that will tell you that the file is encrypted.

dbampalikis (approved):

norling (approved):

encrypt/encrypt.go
```diff
@@ -187,6 +193,27 @@ func checkFiles(files []helpers.EncryptionFileSet) error {
...
+			}
+		}()
+
+		// Extracting the first 8 bytes of the header - crypt4gh
+		byteSlice := make([]byte, 8)
+		magicWord, err := unEncryptedFile.Read(byteSlice)
+		if err != nil {
+			return err
+		}
+		if string(byteSlice[0:magicWord]) == "crypt4gh" {
```
aaperis, line 214: no need to index `byteSlice`

```suggestion
		if string(byteSlice) == "crypt4gh" {
```

aaperis (approved):
it works well, some aesthetic comments that it's up to you consider  :-)

encrypt/encrypt.go
shreyasshivakumara: Ok

PR #42, opened 2022-03-23
Enhancement/continue file encryption after checks

### LINKED ISSUE #11: do not exit if an encrypted file already exists but check and continue
if a file from the input list already has a corresponding encrypted file,
- check that the later is encrypted with the same pk as given in the input
- if yes, skip it and continue with the encryption of the next file from the input list

aaperis:
This PR adds the boolean flag `-continue` to encrypt. If `-continue=true`, encrypt will process all input files that pass the checks but ignore those that do not, instead of exiting prematurely. Info on the files skipped is returned to the user in the output logs.

Closes #11

### DIFF
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -68,6 +68,17 @@ func Encrypt(args []string) error {
 	// All filenames that pass the checks are read into a struct together with their output filenames
 	files := []helpers.EncryptionFileSet{}
 
+	// Counter for skipped files
+	skippedFiles := 0
+
+	// Make sure to exit with error status if any file is skipped
+	defer func() {
+		if skippedFiles != 0 {
+			log.Errorf("(%d/%d) files skipped", skippedFiles, len(files)+skippedFiles)
+			os.Exit(1)
+		}
+	}()
+
 	log.Info("Checking files")
 	for _, filename := range Args.Args() {
 
@@ -80,9 +91,10 @@ func Encrypt(args []string) error {
 
 		eachFile[0] = helpers.EncryptionFileSet{Unencrypted: filename, Encrypted: outFilename}
 
-		// Skip files that do not pass the checks
+		// Skip files that do not pass the checks and print all error logs at the end
 		if err = checkFiles(eachFile); err != nil {
-			log.Errorf("Skipping input file %s. Reason: %s.", filename, err)
+			defer log.Errorf("Skipping input file %s. Reason: %s.", filename, err)
+			skippedFiles++
 
 			continue
 		}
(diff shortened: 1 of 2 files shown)


shreyasshivakumara (approved):
Looks good and thanks for making my changes as well. :)

norling (changes requested):
The code looks good, except that error messages should not be capitalized (as per https://staticcheck.io/docs/checks#ST1005).
I'm sort of torn on this functionality though. I feel like having the script continue without confirmation when there are errors in the input will make it much easier to fail, while not really adding much convenience... If this tool is built into a workflow it'll be easy to miss that not all files are encrypted.

dbampalikis (approved):

norling (approved):

encrypt/encrypt.go
```diff
@@ -206,13 +215,13 @@ func checkFiles(files []helpers.EncryptionFileSet) error {
...
+		magicWord := make([]byte, 8)
+		_, err = unEncryptedFile.Read(magicWord)
 		if err != nil {
-			return err
+			return fmt.Errorf("error reading input file %s, reason: %v", file.Unencrypted, err)
 		}
-		if string(byteSlice[0:magicWord]) == "crypt4gh" {
-			return fmt.Errorf("Input file %s is already encrypted(.c4gh)", file.Unencrypted)
+		if string(magicWord) == "crypt4gh" {
+			return fmt.Errorf("input file %s is already encrypted(.c4gh) - make sure the right pk was used", file.Unencrypted)
```
jbygdell, line 224:
```suggestion
			return fmt.Errorf("input file %s is already encrypted(.c4gh)", file.Unencrypted)
```
If the file already is encrypted wrong input file was selected, not the wrong public key.

encrypt/encrypt_test.go
```diff
@@ -115,7 +115,7 @@ func (suite *EncryptTests) TestcheckFiles() {
 	// Encrypted file is given as input
 	verifyUnencrypted := helpers.EncryptionFileSet{Unencrypted: suite.encryptedFile.Name(), Encrypted: "does-not-exist"}
 	err = checkFiles([]helpers.EncryptionFileSet{verifyUnencrypted})
-	assert.EqualError(suite.T(), err, fmt.Sprintf("Input file %s is already encrypted(.c4gh)", suite.encryptedFile.Name()))
+	assert.EqualError(suite.T(), err, fmt.Sprintf("input file %s is already encrypted(.c4gh) - make sure the right pk was used", suite.encryptedFile.Name()))
```
jbygdell, line 118:
```suggestion
	assert.EqualError(suite.T(), err, fmt.Sprintf("input file %s is already encrypted(.c4gh)", suite.encryptedFile.Name()))
```

jbygdell (changes requested):

encrypt/encrypt.go
aaperis: That was meant to be a side note to the user, i.e. since we cannot check if this file was previously encrypted with the correct pk. But this looks confusing so I removed it.

norling (approved):

jbygdell (approved):

PR #68, opened 2022-05-24
Feature/release

dbampalikis:
This PR enables the automatic creation of releases when a new tag is created, through a github actions workflow.

Summary of changes:
- Added goreleaser workflow, which runs every time a new tag is created
- Added instructions for creating a release


.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+      matrix:
+        go-version: [1.17]
+    steps:
+
+      - name: Set up Go  ${{ matrix.go-version }}
+        uses: actions/setup-go@v3
+        with:
+          go-version:  ${{ matrix.go-version }}
+        id: go
```
jbygdell, line 20:
```suggestion
    steps:

      - name: Set up Go  ${{ matrix.go-version }}
        uses: actions/setup-go@v3
        with:
          go-version:  1.18
        id: go
```

There is not really any point in having a strategy with only one GO version.
And GO 1.17 will be EOL in early September when 1.19 is released so it is better to user 1.18

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+              dep ensure
+          fi
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2.9.1
+        with:
+          version: latest
+          args: release --rm-dist
+          key: ${{ secrets.YOUR_PRIVATE_KEY }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jbygdell, line 39: Maybe also use `.goreleaser.yaml` so we can skip the `i386` and `arm` versions.

.goreleaser.yaml
```diff
@@ -0,0 +1,30 @@
+# This is an example .goreleaser.yml file with some sensible defaults.
```
norling, line 1: It's not important, but I'd prefer to remove comments like this for something more descriptive to this specific file.

README.md
```diff
@@ -133,3 +133,7 @@ Given that the instructions in the [download section](#download) have been follo
 ```
 where `<keypair_name>.sec.pem` the private key created in the [relevant section](#create-keys) and `<file_to_decrypt>` one of the files downloaded following the instructions of the [download section](#download-file).
 
+
+# Create new release
+
+The github actions include a release workflow that builds binaries for different operating systems. In order to create a new release, create a tag either using the graphical interface or through the command line. That should trigger the creation of a release with the latest code of the specified branch.
```
norling, line 139: If I understand the `release.yml` correctly, releases are only created on tags that match `v*` right? That should probably be metioned here if that's the case :)

norling (approved):
Some minor stuff, but otherwise it looks nice! :)

aaperis (approved):
I agree with previous comments, otherwise it looks good :-)

PR #75, opened 2022-06-02
Feature/upload to different path

### LINKED ISSUE #61: Upload to a different path
path/to/file.c4gh new_path/ -> new_path/file.c4gh

`follow standard cp behavior`

aaperis:
This PR adds the functionality to specify a new target folder (or path) e.g. `path/to/file.c4gh new_path/ -> new_path/file.c4gh` when uploading files or folders. 
Summary of changes:
- adds a new flag `-targetDir` with argument a folder name or path
- updates unit and integration tests
- updates readme with instructions and usage examples
- upon request, `-targetDir` can also be passed after the flagless arguments, i.e. at the end of the command
- in fact the above behavior is generalized; any flag should now be possible to be passed either before or after the flagless arguments (seemingly by-passing go's standard behavior)

Closes #61

Note: This PR is based on #73 and merges into that branch, for now at least.


upload/upload.go
```diff
@@ -233,12 +235,35 @@ func Upload(args []string) error {
...
+	// Dereference the pointer to a string
+	uploadDirString := ""
+	if uploadDir != nil {
+		uploadDirString = *uploadDir
+	}
+
+	if (!os.IsNotExist(err) && !info.IsDir()) || (uploadDirString != "" && uploadDirString[0:1] == "-") {
+		return errors.New(*uploadDir + " is not a valid upload directory")
+	}
+
```
jbygdell, line 265: Nitpicking here, but isn't source directory the proper term?

upload/upload.go
```diff
@@ -132,7 +134,7 @@ func checkTokenExpiration(accessToken string) (bool, error) {
 }
 
 // Function uploadFiles uploads the files in the input list to the s3 bucket
-func uploadFiles(files, outFiles []string, config *Config) error {
+func uploadFiles(files, outFiles []string, outDir string, config *Config) error {
```
jbygdell, line 137: Shouldn't `outDir` be `targetDir` also in the code?

jbygdell (approved):
Apart from some nitpicking on naming it looks good.
Although I would like to see proper tests for the actual functionality as well.

norling (approved):

upload/upload.go
aaperis: The above part refers to the `target` directory (though I now see how this can be confusing). I changed `upload` to `target` whenever possible so as to make var names and text clearer.

upload/upload.go
aaperis: Yes, done.

aaperis:
> Although I would like to see proper tests for the actual functionality as well.

Good point! Functionality tests with fakeS3 are now added in the upload test-suite.

dbampalikis (approved):

PR #76, opened 2022-06-03
Feature/list

dbampalikis:
This PR implements the list functionality for the cli.


list/list.go
```diff
@@ -0,0 +1,124 @@
...
+			switch aerr.Code() {
+			case s3.ErrCodeNoSuchBucket:
+				fmt.Println(s3.ErrCodeNoSuchBucket, aerr.Error())
+			default:
+				fmt.Println(aerr.Error())
+			}
+		} else {
+			// Print the error, cast err to awserr.Error to get the Code and
+			// Message from an error.
+			fmt.Println(err.Error())
```
norling, line 70: Shouldn't the error be returned and printed from main, or printed with `log.Error`? Won't the message go to `stdout` instead of `stderr` if you just to `fmt.Println`?

norling (approved):
I'm a bit unclear on some of the error handling, but looks good otherwise! :)

.github/integration/tests/tests.sh
```diff
@@ -50,6 +50,15 @@ check_encypted_file $files
 ./sda-cli upload -config sda-s3proxy/dev_utils/s3cmd.conf data_file.c4gh
 check_uploaded_file test/dummy/data_file.c4gh data_file.c4gh
 
+echo "Here"
```
jbygdell, line 53:
```suggestion
```

list/list.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		return nil, err
+	}
```
jbygdell, line 74: This needs to be reworked, as it is the modified message is printed to sdtout but the original error is what's returned to main.

list/list.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Check that the s3 configuration file path exists
+	if *configPath == "" {
+		return errors.New("failed to find an s3 configuration file for listing data")
+	}
+
+	// Get the configuration in the struct
+	config, err := upload.LoadConfigFile(*configPath)
+	if err != nil {
+		return err
```
jbygdell, line 102: This assumes that the called function returns a proper error message

list/list_test.go
```diff
@@ -0,0 +1,37 @@
...
+	assert.EqualError(suite.T(), err, "failed to find an s3 configuration file for listing data")
+}
+
+func (suite *TestSuite) TestTooManyArgs() {
+
+	os.Args = []string{"list", "arg1", "arg2"}
+
+	err := List(os.Args)
+	assert.EqualError(suite.T(), err, "failed to parse prefix, only one is allowed")
+}
```
jbygdell, line 37: This doesn't test the actual listing functionality

jbygdell (changes requested):

list/list.go
dbampalikis: I am not sure what you mean with this comment. The `LoadConfigFile` is supposed to return error messages when something is wrong. is there anything more on that?

list/list_test.go
jbygdell, line 37: Still missing actual test for listing files.

jbygdell (changes requested):

PR #88, opened 2022-06-20
Disconnect from S3-proxy repo

jbygdell:


.github/integration/tests/tests.sh
```diff
@@ -47,12 +47,12 @@ check_encypted_file $files
...
 
 # Upload a specific file and check it
-./sda-cli upload -config sda-s3proxy/dev_utils/s3cmd.conf data_file.c4gh
+./sda-cli upload -config testing/s3cmd.conf data_file.c4gh
 check_uploaded_file test/dummy/data_file.c4gh data_file.c4gh
 
 
 # Create and encrypt multiple files in a folder
-dd if=/dev/random of=data_file1 count=1 bs=$(( 1024*1024 ))
+dd if=/dev/random of=data_file1 count=1 bs=1M
```
norling, line 55: should be fine to use `/dev/urandom` here as well right?

norling (approved):

dbampalikis (approved):
Looks good! Could you change the message of the last commit though?

PR #95, opened 2022-07-11
Feature/encrypt and upload

### LINKED ISSUE #60: Automatic encryption on upload
* ensure files are not already encrypted
* in the upload use `--encrypt-with-key` flag to define the key that will be used for encryption

aaperis:
This PR adds "encrypt and upload" functionality to the sda-cli's `upload` command. Specifically, the user can pass the flag `--encrypt-with-key` followed by the public key file and the source files will be first encrypted and then uploaded in one go. The code calls the `Encrypt` function internally so it inherits its behavior for the first step. If input source files include already encrypted files then the tool will exit before encryption and thus before upload.

Main additions:

- add new flag `--encrypt-with-key` and related functionality
- extend test suite of upload
- add integration tests
- update readme

Closes #60

### DIFF
--- a/.github/integration/tests/tests.sh
+++ b/.github/integration/tests/tests.sh
@@ -118,6 +118,33 @@ do
     check_uploaded_file test/dummy/$uploadDir"/"$k $k
 done
 
+# Encrypt and upload
+
+mkdir data_files_unenc && mkdir data_files_unenc/dir1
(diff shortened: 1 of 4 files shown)


norling (approved):
Neat! :)

dbampalikis (approved):

PR #116, opened 2022-10-06
Feature/multiple key encryption

aaperis:
This PR adds multiple key encryption capability to `encrypt`, in summary the following functionality is added:

- input multiple public keys in the command prompt by using the -key flag repeatedly. This behavior mimics that of the crypt4gh tool

- an argument to -key can be a file with concatenated C4GH public keys and `encrypt` will parse and use them for encryption

- encrypt performs an early basic check and discards key files with invalid key format by reading the first few bytes of the given  file. This ensures that any mistakes of passing a wrong file as key file are avoided.

- multiple key encryption is also enabled for the `-encrypt-with-key` feature of the `upload` module  via passing a file with concatenated public keys as argument (and letting `encrypt` work on it in the background). Multiple `-encrypt-with-key` flag calls seemed to me like a superfluous feature for `upload` so it was not implemented for that module by choice.

Some refactoring regarding how the key is read was done and a few new functions were added to implement the feature. The go test suite, integration tests and readme files were also updated accordingly.

Closes [#534](https://github.com/NBISweden/LocalEGA-SE-Deployment/issues/534)

### DIFF
--- a/.github/integration/tests/tests.sh
+++ b/.github/integration/tests/tests.sh
@@ -58,11 +58,10 @@ else
     exit 1
 fi
 
-
 # Create and encrypt multiple files in a folder
-cp data_file data_file1
 
 # Create folder and encrypt files in it
+cp data_file data_file1
 mkdir data_files_enc
 ./sda-cli encrypt -key sda_key.pub.pem -outdir data_files_enc data_file data_file1
 
@@ -76,6 +75,7 @@ do
 done
 
 # Test recursive folder upload
+
 # Create folder with subfolder structure and add some encrypted files
 mkdir data_files_enc/dir1 data_files_enc/dir1/dir2
 cp data_files_enc/data_file.c4gh data_files_enc/data_file3.c4gh
@@ -93,6 +93,7 @@ do
 done
 
 # Test upload to a different path
+
 # Upload a folder recursively and a single file in a specified upload folder
 uploadDir="testfolder"
 ./sda-cli upload -config testing/s3cmd.conf -targetDir "$uploadDir" -r data_files_enc/dir1 data_files_enc/data_file3.c4gh
@@ -108,6 +109,7 @@ do
 done
 
 # Upload all contents of a folder recursively to a specified upload folder
+
 uploadDir="testfolder2"
 ./sda-cli upload -config testing/s3cmd.conf -targetDir "$uploadDir" -r data_files_enc/dir1/.
 
@@ -126,11 +128,7 @@ cp data_file data_files_unenc/. && cp data_file data_files_unenc/dir1/data_file1
 uploadDir="testEncryptUpload"
 ./sda-cli upload -config testing/s3cmd.conf -encrypt-with-key sda_key.pub.pem -r data_files_unenc -targetDir "$uploadDir"
 
-for k in data_files_unenc/data_file.c4gh data_files_unenc/dir1/data_file1.c4gh
-do
-    check_encypted_file $k
-done
-# check_encypted_file "data_files_unenc/data_file.c4gh" "data_files_unenc/dir1/data_file1.c4gh"
+check_encypted_file "data_files_unenc/data_file.c4gh" "data_files_unenc/dir1/data_file1.c4gh"
 
 for k in data_files_unenc/data_file.c4gh data_files_unenc/dir1/data_file1.c4gh
 do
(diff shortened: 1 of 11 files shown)


encrypt/encrypt.go
```diff
@@ -114,9 +127,15 @@ func Encrypt(args []string) error {
 
 	log.Infof("Ready to encrypt %d file(s)", len(files))
 
-	// Read the public key to be used for encryption. The private key
-	// matching this public key will be able to decrypt the file.
-	publicKey, err := readPublicKey(*publicKeyFile)
+	// Initialize a c4gh public key specs instance
+	c4ghKeySpecs := keySpecs{
+		rgx:    regexp.MustCompile(`-{5}[A-Z]*\s[4A-Z]*\s[A-Z]*\s[A-Z]*-{5}\n.*\n-{5}[A-Z]*\s[4A-Z]*\s[A-Z]*\s[A-Z]*-{5}`),
```
norling, line 132: you have these regexes - it might be clearer to just use `-----BEGIN CRYPT4GH PUBLIC KEY-----` as a pattern instead of `-{5}[A-Z]*\s[4A-Z]*\s[A-Z]*\s[A-Z]*-{5}` - or does this need to match other strings too? :)

norling (approved):
The regexes seem like they could be simpler (if I haven't misunderstood them), but otherwise this seems sounds :)

encrypt/encrypt.go
aaperis: my first thought was to allow some verbosity freedom in case the format of the key is changed (for some reason). but I see how this was unnecessary, so I changed it to something more to the point. Thanks for pointing out!

.github/integration/tests/tests.sh
```diff
@@ -173,6 +171,78 @@ else
...
+        echo "Created key pair for encryption"
+    else
+        echo "Failed to create key pair for encryption"
+        exit 1
+fi
+done
+
+# Create file with concatenated pub keys
+cat sda_key1.pub.pem sda_key2.pub.pem > sda_keys
+
```
kostas-kou, line 189: Maybe here move the creation of the concatenated pub keys to the section where you actually encrypt with this concatenated key (line 211)?

kostas-kou (changes requested):

kostas-kou (approved):

encrypt/encrypt_test.go
```diff
@@ -120,11 +141,55 @@ func (suite *EncryptTests) TestcheckFiles() {
...
+
+func (suite *EncryptTests) TestreadPublicKeyFile() {
+	publicKey, err := readPublicKeyFile(suite.publicKey.Name())
 	assert.NoError(suite.T(), err)
 	suite.Equal(*publicKey, suite.pubKeyData)
 }
 
+func (suite *EncryptTests) TestreadMultiPublicKeyFile() {
+	specs := keySpecs{
+		rgx:    regexp.MustCompile(`-{5}[A-Z]*\s[4A-Z]*\s[A-Z]*\s[A-Z]*-{5}\n.*\n-{5}[A-Z]*\s[4A-Z]*\s[A-Z]*\s[A-Z]*-{5}`),
```
jbygdell, line 167:
```suggestion
		rgx:    regexp.MustCompile(`-{5}BEGIN CRYPT4GH PUBLIC KEY-{5}\n.*\n-{5}END CRYPT4GH PUBLIC KEY-{5}`),
```

encrypt/encrypt_test.go
```diff
@@ -120,11 +141,55 @@ func (suite *EncryptTests) TestcheckFiles() {
...
+	publicKey, err := readMultiPublicKeyFile(suite.multiPublicKey.Name(), specs)
+	assert.NoError(suite.T(), err)
+	b := *publicKey
+	suite.Equal(b[0], suite.pubKeyData)
+	suite.Equal(b[1], suite.pubKeyData)
+}
+
+func (suite *EncryptTests) TestcheckKeyFile() {
+	specs := keySpecs{
+		rgx:    regexp.MustCompile(`-{5}[A-Z]*\s[4A-Z]*\s[A-Z]*\s[A-Z]*-{5}\n.*\n-{5}[A-Z]*\s[4A-Z]*\s[A-Z]*\s[A-Z]*-{5}`),
```
jbygdell, line 179:
```suggestion
		rgx:    regexp.MustCompile(`-{5}BEGIN CRYPT4GH PUBLIC KEY-{5}\n.*\n-{5}END CRYPT4GH PUBLIC KEY-{5}`),
```

jbygdell (approved):
To make it easier with the regexp you can make a function like so:
```
func newKeySpecs() keySpecs{} {
	return keySpecs{
		rgx:    regexp.MustCompile(`-{5}BEGIN CRYPT4GH PUBLIC KEY-{5}\n.*\n-{5}END CRYPT4GH PUBLIC KEY-{5}`),
		nbytes: 115,
	}
```
Then it only needs to be updated in one place and it can be reused for the tests

aaperis:
> To make it easier with the regexp you can make a function like so:
> 
> ```
> func newKeySpecs() keySpecs{} {
> 	return keySpecs{
> 		rgx:    regexp.MustCompile(`-{5}BEGIN CRYPT4GH PUBLIC KEY-{5}\n.*\n-{5}END CRYPT4GH PUBLIC KEY-{5}`),
> 		nbytes: 115,
> 	}
> ```
> 
> Then it only needs to be updated in one place and it can be reused for the tests

Great suggestion, thanks! It is now done like that.

PR #153, opened 2023-01-24
Fix releaser bug on multiple archives

dbampalikis:
The pull request if fixing the issue with the previous PR, which broke the releases, and solves the issue with creating `zip` instead of `tar` files for windows users. 

Also includes a small README update on checking the configuration file for the releaser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,3 +221,9 @@ The github actions include a release workflow that builds binaries for different
 
 In order for the automatic release to get triggered, the releases should be of the format `vX.X.X`, e.g. `v1.0.0`.
 
+## Update releaser
+
+Before pushing a change to the releaser, make sure to run the check for the configuration file, running:
+```sh
+goreleaser check -f .goreleaser.yaml
+```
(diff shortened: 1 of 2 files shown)


aaperis (approved):

README.md
```diff
@@ -221,3 +221,9 @@ The github actions include a release workflow that builds binaries for different
 
 In order for the automatic release to get triggered, the releases should be of the format `vX.X.X`, e.g. `v1.0.0`.
 
+## Update releaser
+
+Before pushing a change to the releaser, make sure to run the check for the configuration file, running:
+```sh
+goreleaser check -f .goreleaser.yaml
```
jbygdell, line 228: I would have put this as a step as a GH-action triggered on `push`  or a similar way so that there is not a prerequisite to have the goreleaser binary installed locally.

dbampalikis: It was a decision to make it manual, so that we only get new version when we want it. Of course we could add specific cases, on when to release. However, I think it is a problem to have the goreleaser installed, we do the same with all github actions linters, so I think that should be fine. Plus it doesn't change that often

norling (approved):

Issue #188, opened 2023-04-14
Fix MultipartChunkSizeMb in `upload.go` 


jbygdell:
This is the current code and it is not logical.
```
if config.MultipartChunkSizeMb <= 15 {
	config.MultipartChunkSizeMb = 50
}
```

Issue #187, opened 2023-04-14
Sanitize paths for `-targetDir `


jbygdell:
Currently the `\` -> `/` conversion is not being done there.

PR #189, opened 2023-04-14
Fix latest batch of user errors

jbygdell:
* Converts windows pat to unix when using the `-targetDir` option, tests on windows by using `filepath.join()`
* sets correct MultipartChunkSizeMb if set below 15MB
* Fixes some linter issues
* print to stdout instead of log when listing uploads

Fixes: #188 #187

### DIFF
--- a/upload/upload_test.go
+++ b/upload/upload_test.go
@@ -95,8 +95,7 @@ access_key = someUser
 
 	defer os.Remove(configPath.Name())
 
-	err = os.WriteFile(configPath.Name(), []byte(confFile), 0600)
-	if err != nil {
+	if err := os.WriteFile(configPath.Name(), []byte(confFile), 0600); err != nil {
 		log.Printf("failed to write temp config file, %v", err)
 	}
 
@@ -170,18 +169,15 @@ func (suite *TestSuite) TestSampleNoFiles() {
 	// Test Upload function
 	os.Args = []string{"upload", "-config", configPath.Name()}
 
-	err = Upload(os.Args)
-	assert.EqualError(suite.T(), err, "no files to upload")
+	assert.EqualError(suite.T(), Upload(os.Args), "no files to upload")
 
 	// Test handling of mistakenly passing a filename as an upload folder
 	os.Args = []string{"upload", "-config", configPath.Name(), "-targetDir", configPath.Name()}
-	err = Upload(os.Args)
-	assert.EqualError(suite.T(), err, configPath.Name()+" is not a valid target directory")
+	assert.EqualError(suite.T(), Upload(os.Args), configPath.Name()+" is not a valid target directory")
 
 	// Test handling of mistakenly passing a flag as an upload folder
 	os.Args = []string{"upload", "-config", configPath.Name(), "-targetDir", "-r"}
-	err = Upload(os.Args)
-	assert.EqualError(suite.T(), err, "-r"+" is not a valid target directory")
+	assert.EqualError(suite.T(), Upload(os.Args), "-r"+" is not a valid target directory")
 
 	// Test passing flags at the end as well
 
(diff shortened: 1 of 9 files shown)


upload/upload_test.go
```diff
@@ -361,11 +355,11 @@ func (suite *TestSuite) TestFunctionality() {
 
 	// Test recursive upload
 	os.Args = []string{"upload", "-config", configPath.Name(), "-r", dir}
-	err = Upload(os.Args)
-	assert.NoError(suite.T(), err)
+	assert.NoError(suite.T(), Upload(os.Args))
 
 	// Check logs that file was uploaded
 	logMsg := fmt.Sprintf("%v", strings.TrimSuffix(str.String(), "\n"))
+	// fmt.Printf("FOO: %s\n", logMsg)
```
norling, line 362: probably forgot this line here
```suggestion
```

upload/upload_test.go
```diff
@@ -379,13 +373,14 @@ func (suite *TestSuite) TestFunctionality() {
 	assert.Equal(suite.T(), aws.StringValue(result.Contents[0].Key), fmt.Sprintf("%s/%s", filepath.Base(dir), filepath.Base(testfile.Name())))
 
 	// Test upload to a different folder
-	os.Args = []string{"upload", "-config", configPath.Name(), testfile.Name(), "-targetDir", "a"}
-	err = Upload(os.Args)
-	assert.NoError(suite.T(), err)
-
+	targetPath := filepath.Join("a", "b", "c")
+	// fmt.Printf("target path: %s\n", targetPath)
```
norling, line 377: more old debug lines
```suggestion
```

upload/upload_test.go
```diff
@@ -379,13 +373,14 @@ func (suite *TestSuite) TestFunctionality() {
...
+	os.Args = []string{"upload", "-config", configPath.Name(), testfile.Name(), "-targetDir", targetPath}
+	assert.NoError(suite.T(), Upload(os.Args))
+	// fmt.Printf("\n\nBAR: %v\n", strings.TrimSuffix(str.String(), "\n"))
```
norling, line 380:
```suggestion
```

upload/upload_test.go
```diff
@@ -395,7 +390,8 @@ func (suite *TestSuite) TestFunctionality() {
 	if err != nil {
 		log.Panic(err.Error())
 	}
-	assert.Equal(suite.T(), aws.StringValue(result.Contents[0].Key), fmt.Sprintf("a/%s", filepath.Base(testfile.Name())))
+	// fmt.Printf("%s/%s\n", filepath.ToSlash(targetPath), filepath.Base(testfile.Name()))
```
norling, line 393:
```suggestion
```

norling (approved):
some forgotten debug printouts, otherwise fine as far as I can tell! :)

upload/upload_test.go
```diff
@@ -471,4 +458,103 @@ func (suite *TestSuite) TestFunctionality() {
...
+	assert.Contains(suite.T(), logMsg, msg)
+
+	// Check that file showed up in the s3 bucket correctly
+	result, err := s3Client.ListObjects(&s3.ListObjectsInput{
+		Bucket: aws.String("dummy"),
+	})
+	if err != nil {
+		log.Printf(err.Error())
+	}
+	assert.Equal(suite.T(), filepath.ToSlash(filepath.Join(targetPath, filepath.Base(dir), filepath.Base(testfile.Name()))), aws.StringValue(result.Contents[0].Key))
```
aaperis, line 557: No need for a new function, you can add this part before line 392 and then increase the index in line 428 to 2 for that test to pass.

aaperis (approved):
There is some code repetition in the added test which I think can be avoided, but otherwise it looks good :-)

upload/upload_test.go
jbygdell: The whole point of having the tests independent is to track failures without having the cascading effects of testing everything together.

PR #193, opened 2023-04-21
allow parsing simple s3cmd config files

### LINKED ISSUE #182: Errors when s3cfg contains headers
If the s3cfg file contains headers `[HEADER]` the sda-cli can't parse the file correctly and thus fails to get the credentials.

A/C
make the sda-cli ignore a line in the config file if it is a header line

aaperis:
This fix allows for s3cmd conf files to be parsed correctly by `sda-cli` if the file has none or one header. 

By default the `go-ini` library used to parse the s3conf file always assumes a `[DEFAULT]` header even when the file has none. In case the file itself  has (for any reason) more than one headers (i.e. `ini` sections), `sda-cli` will parse only the section that corresponds to the first one.

Closes #182.

### DIFF
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -77,7 +77,15 @@ func LoadConfigFile(path string) (*Config, error) {
 		return config, err
 	}
 
-	if err := cfg.Section("").MapTo(config); err != nil {
+	// ini sees a DEFAULT section by default
+	var iniSection string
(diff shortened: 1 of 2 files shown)


jbygdell (approved):

kusalananda (approved):
This seems to resolve our issue with the sections in the configuration file.

PR #207, opened 2023-05-15
#204: clarify wildcard user for recursive uploads

### LINKED ISSUE #204: Clarify wildcard use for recursive uploads
A comment from the external user testing was:
> For recursive uploading, it is unclear which wildcards should be used, . or *.

It should be clarified in text and examples how wildcards can be used.

pahatz:
This looks like an insignificant change, but since it was pointed out, we can change it to make the example clearer.

closes #204


norling (approved):
Good Job! :)

kusalananda (changes requested):
Should probably talk about how a dot matters at the end of a directory path instead of changing the dot to a star.  See conversation in associated issue, #204

kusalananda (approved):
Good

Issue #203, opened 2023-05-12
Progress indicator for uploads


norling:
There is currently no feedback to the user while uploading files, which is quite unnerving when large uploads take a long time.
Ideally a progress indicator containing current upload speed and upload progress should be added, but at a minimum an indicator the upload is still progressing should be added.

PR #208, opened 2023-05-16
#203: add upload progress indicator

pahatz:
closes: #203 

Add an indicator for the upload so the user knows it is still active.
The indicator should be showing a progress bar for each individual file and the upload speed.


upload/upload.go
```diff
@@ -144,6 +147,45 @@ func CheckTokenExpiration(accessToken string) (bool, error) {
 	return tomorrow.After(expiration), nil
 }
 
+// progress bar stuff
```
norling, line 150: I'd like some more documentation about what this stuff actually does! 
Also - in case we wanna add a progress bar to download as well, we might wanna have the CustomReader in helpers (or it's own file).

norling (approved):
So pretty! :)

jbygdell (changes requested):

upload/upload.go
pahatz: I'll move it in helpers for a start and add a better description.

nanjiangshu (commented):
Great work @pahatz! The progress indicator looks nice. However, I find two unexpected behaviors when test uploading to BP on Mac.
1. After uploading of one file is finished, the terminal prints out a line like 
```&{https://inbox.bp.nbis.se/1a8755401d7d26ef96966e2e4da7c826cb905c60_elixir-europe.org/dummy.txt.c4gh <nil>  0xc0003a0130}```
I don't think this line needs to be printed since at next line it will print the INFO message 
``` INFO[0001] file uploaded to https://inbox.bp.nbis.se/1a8755401d7d26ef96966e2e4da7c826cb905c60_elixir-europe.org/dummy.txt.c4gh ```
Not to say messages like `<nil>  0xc0003a0130}` is not proper for the end-user

2. When uploading multiple files, it seems the indicator bar of the previously uploaded files will be printed as well, which results many duplicated messages. E.g. 
```File#1: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#2: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
&{https://inbox.bp.nbis.se/1a8755401d7d26ef96966e2e4da7c826cb905c60_elixir-europe.org/test_dataset_big/temp_100m_file_10.c4gh <nil> 2~RGiemOAmjRyIbUjrt_ygwItsV1DcxVb 0xc0000f1408}
File#1: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#2: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#3: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#1: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#2: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#3: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#1: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#2: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#3: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#1: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#2: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
File#3: uploading 100.0 MiB / 100.0 MiB [========================================================================================================================================================================================================================]  done
```

upload/upload.go
```diff
@@ -175,9 +218,31 @@ func uploadFiles(files, outFiles []string, targetDir string, config *Config) err
 			return err
 		}
 
+		fileInfo, err := f.Stat()
+
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
```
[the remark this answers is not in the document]
pahatz: I included all your suggestions.

upload/upload.go
pahatz: It's moved, Some more comments added too.

nanjiangshu (approved):
Awesome!

jbygdell (approved):

Issue #199, opened 2023-05-12
Make argument position irrelevant


norling:

pahatz:
If I'm not mistaken, it's the opposite. This works:
`./sda-cli download url_list.txt -c testing/s3cmd.conf -outdir bla `
while this doesn't: 
`./sda-cli download -c testing/s3cmd.conf -outdir bla url_list.txt`

pahatz:
Yes, that would also be a good solution. Honestly, I don't see this as a big issue since the documentation is clear on the order they should come. In the PR's state right now, it should just be able to handle the wrong in the rest of the modules.

PR #214, opened 2023-05-23
#199: make argument position irrelevant

pahatz:
Make the order of the flags and arguments irrelevant when using the modules decrypt, download, encrypt, list.
The code from the upload module that handled this previously was moved to helpers.


jbygdell:
If the file with the urls where matched with a `-f` flag, the order in which they appear would be irrelevant.

norling (approved):

dbampalikis (approved):
Works well. Good job

Issue #200, opened 2023-05-12
Warning on overwriting remote file


norling:
There is currently no feedback from the system whether a file exists in the remote when attempting to upload. This should be added, optionally with a `-f/--force` flag to automatically overwrite remote files.

The required refactoring to resolve the circular dependencies will also be part of the scope on this.
config should be move out of upload and into a new package.

jbygdell:
This will be complicated to implement.
Mainly because the S3 backend doesn't care if you overwrite an existing file or not.

PR #215, opened 2023-05-25
#200: warn when overwriting a files that already exists in the bucket

pahatz:
Warn if the user attempts to overwrite a file that already exists in the bucket.


pahatz:
Needs some refactoring and test fixing

kusalananda (commented):
Ok, so here you have a force option (`-f`), which mean sit would not be appropriate to use the same option when forcing the uploading unencrypted files (another PR).

However, one could use `-force overwrite` and `-force unencrypted` (or similar option-argument names) and `-force all` (for both).

upload/upload.go
```diff
@@ -175,6 +80,28 @@ func uploadFiles(files, outFiles []string, targetDir string, config *Config) err
...
+		files, err := helpers.ListFiles(*config, targetDir)
+		if err != nil {
+			log.Error("Couldn't get the file list ", err)
+		}
+		if len(files.Contents) > 0 {
+
+			// range through Contents and check if we have a match
+			for l := range files.Contents {
+				// check if the two strings match
+				if aws.StringValue(files.Contents[l].Key) == config.AccessKey+"/"+targetDir+"/"+filepath.Base(filename) {
```
aaperis, line 93:
```suggestion
				if aws.StringValue(files.Contents[l].Key) == config.AccessKey+path.Clean("/"+targetDir+"/")+outFiles[k] {
```

If `targetDir` is empty then there's a "//" that needs to be taken care of. 
Use of `outFiles[k]` ensures that the correct upload path of the file is considered when e.g. recursive folder upload is done.

pahatz: I added it! Also removed the loop as you suggested, only listing for each specific filename.

pahatz:
> Ok, so here you have a force option (`-f`), which means it would not be appropriate to use the same option when forcing the uploading of unencrypted files (another PR).
> 
> However, one could use `-force overwrite` and `-force unencrypted` (or similar option-argument names) and `-force all` (for both).

Changed here too to use --force-overwrite

aaperis (commented):
It looks pretty cool and it works great. Some suggestions for things to refactor :-)

aaperis:
While at it, there are three  `fmt.Println`'s in this function:
https://github.com/NBISweden/sda-cli/blob/3766708c5dc7b79ab9ce4f0b315905111dd7bd86/helpers/helpers.go#L141
that need to be removed :-)

pahatz:
This PR currently resolves a few more issues that what was initially intended to:

- Introduce a --force-overwrite flag that is used when we want to overwrite existing files.
- Check in upload before uploading if the file already exists. Some cases to handle differently between directories and single files when checking.
- Move ListFiles func in helpers to be able to use it in upload. That resulted in a lot of refactoring to sort out the circular dependencies that came up. The corresponding tests where also moved.
- Flag parsing for upload module now uses helpers.ParseArgs func as well. Fixed the logic for that func.

aaperis (approved):
This looks and works great! :-)

PR #216, opened 2023-05-25
201: warn when uploading an unencrypted file

pahatz:
Issue a warning when the user is trying to upload a file that is not encrypted.
We check the firt 8 bytes of the header which according to crypt4gh specification doc
is a "magicWord" equal to "crypt4gh"


upload/upload.go
```diff
@@ -174,6 +174,17 @@ func uploadFiles(files, outFiles []string, targetDir string, config *Config) err
...
 		}
+		// Check if the file is encrypted and warn if not
+		// Extracting the first 8 bytes of the header - crypt4gh
+		magicWord := make([]byte, 8)
+		_, err = f.Read(magicWord)
+		if err != nil {
+			fmt.Printf("error reading input file %s, reason: %v", filename, err)
+		}
+		if string(magicWord) != "crypt4gh" {
+			fmt.Printf("input file %s is not encrypted", filename)
```
norling, line 185: I'd prefer that all files are checked before any uploads start, and that upload (of anything) only starts if something like the `continue` flag from the encrypt module is given. Otherwise there's a risk that a lot of partial unencrypted files are uploaded! :)

norling (changes requested):

kusalananda (commented):
I have two issues or questions about this:

1. When a file is found to be unencrypted, using `-continue` will still upload that file, right? Or will it _continue_ with the next file? If it uploads the unencrypted files, then I think a better name for the option would be `-force`, which (IMHO) better implies that we are overriding the default behaviour of not uploading unencrypted files.

2. Are "long" options supposed to have a single dash (`-config`) or double dash (`--encrypt-with-key`)?

pahatz:
> I have two issues or questions about this:
> 
> 1. When a file is found to be unencrypted, using `-continue` will still upload that file, right? Or will it _continue_ with the next file? If it uploads the unencrypted files, then I think a better name for the option would be `-force`, which (IMHO) better implies that we are overriding the default behaviour of not uploading unencrypted files.  EDIT: Or `-force unencrypted` (see my comment to the PR about overwriting already uploaded files).
> 2. Are "long" options supposed to have a single dash (`-config`) or double dash (`--encrypt-with-key`)?

Good point, as you mentioned in the other PR too. I think it's better we go with this suggestion.
On the second point, I think the default golang command line argument parser is a bit limited but it can handle long arguments.

dbampalikis (approved):
Looks good! I like Andreas comment, but otherwise tested and works well!

aaperis:
looks good to me, the problem with boolean flags is that their position is not independent. For example:
`./sda-cli upload -config testing/s3cmd.conf -continue data_files_unenc -r`
won't work

pahatz:
> looks good to me, the problem with boolean flags is that their position is not independent. For example: `./sda-cli upload -config testing/s3cmd.conf -continue data_files_unenc -r` won't work, whereas `./sda-cli upload -config testing/s3cmd.conf -continue -r data_files_unenc` works.

Did you miss something in the first case? Isn't it -r <some_dir>

kusalananda:
> > I have two issues or questions about this:
> > 
> > 1. When a file is found to be unencrypted, using `-continue` will still upload that file, right? Or will it _continue_ with the next file? If it uploads the unencrypted files, then I think a better name for the option would be `-force`, which (IMHO) better implies that we are overriding the default behaviour of not uploading unencrypted files.  EDIT: Or `-force unencrypted` (see my comment to the PR about overwriting already uploaded files).
> > 2. Are "long" options supposed to have a single dash (`-config`) or double dash (`--encrypt-with-key`)?
> 
> Good point, as you mentioned in the other PR too. I think it's better we go with this suggestion. On the second point, I think the default golang command line argument parser is a bit limited but it can handle long arguments.

Currently, it seems the built-in usage text uses long options with a mix of both `-` and `--`.

aaperis (approved):

Issue #205, opened 2023-05-12
Improve user messages


norling:
Currently many user facing messages look like server log output. This should be changed so that tools have the "normal unix look and feel".

PR #218, opened 2023-06-02
#205: Improve user messages

pahatz:
Switch messages that are destinated for the user from log to fmt.
Set level for logs to avoid having any low level logs reaching the user.


pahatz:
A lot of uni tests rely on comparing against the logs to assert. The changes break those.

upload/upload.go
```diff
@@ -168,7 +168,8 @@ func uploadFiles(files, outFiles []string, targetDir string, config *Config) err
 
 	for k, filename := range files {
 
-		log.Infof("Uploading %s with config %s", filename, *configPath)
+		log.Infof("Uploading %s with config %s\n", filename, *configPath)
```
aaperis, line 171:
```suggestion
```
right? see also below

upload/upload.go
```diff
@@ -189,7 +190,8 @@ func uploadFiles(files, outFiles []string, targetDir string, config *Config) err
 		if err != nil {
 			return err
 		}
-		log.Infof("file uploaded to %s", string(aws.StringValue(&result.Location)))
+		log.Infof("file uploaded to %s\n", string(aws.StringValue(&result.Location)))
```
aaperis, line 193:
```suggestion
```

pahatz: Some unit tests are using this log to assert the result. I left these with level=Info so they are only available if we raise the loglevel, debugging purposes etc.

norling (approved):

create_key/create_key.go
```diff
@@ -102,7 +102,7 @@ func GenerateKeyPair(basename, password string) error {
 	}
 	defer func() {
 		if err := pubFile.Close(); err != nil {
-			log.Errorf("Error closing file: %s\n", err)
+			fmt.Printf("Error closing file: %s\n", err)
```
kusalananda, line 105: Errors should go to the standard error stream.

```suggestion
			fmt.Fprintf(os.Stderr, "Error closing file: %v\n", err)
```

Note change to `%v` too. It may be worth while investigating whether one should use `%w` to format the error string.

Note: This also applies to all other changes from `log.Error()` to `fmt.Printf` in this PR.

kusalananda (changes requested):
See comment (errors and warnings to standard error stream, not standard output).

jbygdell (changes requested):
Keep `log.Error` for real errors and just use `fmt.Print` for messages intended to be displayed to the user if no error occurs.

pahatz:
> Keep `log.Error` for real errors and just use `fmt.Print` for messages intended to be displayed to the user if no error occurs.



> Keep `log.Error` for real errors and just use `fmt.Print` for messages intended to be displayed to the user if no error occurs.

I think the error messages reaching the user should also not have the logrus format that currently looks like: 
FATA[0000] no files to upload 
 

I switched back to using log.Error for the rest but I might have missed some, so please double check

upload/upload.go
pahatz: Both tests and "log not reaching the user" seem to work, so it should be fine like this

aaperis (approved):
Looks good to me :-)

pahatz:
@jbygdell @kusalananda Could you take another look now?

PR #234, opened 2023-06-28
#227 login from cli

### LINKED ISSUE #227: Handle login within cli instead of using a config with the token
Currently the cli is does not handle auth directly but the user needs to provide the config with their access token as an argument.
It would be a big improvement to be able to login directly from the cli by providing credentials and target url/clientID

This ticket should cover the basics to achieve the login by fetching first from auth endpoint /info the data:
```{
        ClientID
	OidcURI 
	PublicKey
	InboxURI
}
```

Then we give the user the lin kto OIDC where they complete the login. 
We can now update the config file with the user's details:

```access_key 
secret_key 
access_token
host_bucket
host_base 
multipart_chunk_size_mb 
guess_mime_type
check_ssl_certificate
encoding
check_ssl_hostname
use_https 
socket_timeout
human_readable_sizes
```

Miro description:
Login to the LocalEGA service for upload using the OIDC "device authentication flow".

Our LifeScience AAI configuration should already be in place.

Upon successful login the configuration parameters should be saved in a .sda-cli-session file (or similar) that can then be used on following usage of sda-cli (such as upload/encrypt and so on). Warn before overwriting.

LS AAI Service ID needs to be in sda-cli or in some configuration file for this to work. An endpoint where this can be reached so it's easy to detect what system (BigPicture or EGA or whatever) is used. Use sda-auth if only the LS AAI client id, otherwise the api service in sda-pipeline (or merged repo).

Endpoint should return everything needed for creating the config file and the following:
Client id
URL for s3 inbox
Public key of repository

## This doesn't complete the login work, will follow up with tickets for:

- session temp file and logic for handling between config and session file [#236](https://app.zenhub.com/workspaces/giant-sloth-6041ec3fd9297d0016321f73/issues/gh/nbisweden/sda-cli/236)
- use the public key provided by login on encrypt [#240](https://app.zenhub.com/workspaces/giant-sloth-6041ec3fd9297d0016321f73/issues/gh/nbisweden/sda-cli/240)
- ~use inboxURI by login in upload~ (already implemented since the session file has the same format as the config file and we fallback on that)
- unit tests/ update README

pahatz:
Login from the cli by providing a URL
The user can run the login <target_URL> command. 
The tool then visits the  <target_URL>/info (provided by auth) where it fetches some parameters needed to complete the login.
The cli then opens the OIDC url where they have to complete the login.
After that the user gets a confirmation and the tool creates a config file in the same directory as the tool with the name of `sda-cli-session`.

closes #227


login/login.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+	if len(Args.Args()) == 1 {
+		url = Args.Args()[0]
+	}
+	log.Println("url: ", url)
+	info, err := GetAuthInfo(url)
+	info.ClientID = "8b7b0168-6b16-4fd2-baec-b0a28b0d5cb0"
+	info.InboxURI = "s3.bp.nbis.se"
+	info.OidcURI = "https://login.elixir-czech.org/oidc"
+	if err != nil {
```
jbygdell, line 179:
```suggestion
	info, err := GetAuthInfo(url)
	if err != nil {
```
These should not be hardcoded.

login/login.go
```diff
@@ -0,0 +1,390 @@
...
+	var userinfo *UserInfo
+	err = json.Unmarshal(body, &userinfo)
+
+	return userinfo, err
+}
+
+// getWellKnown() makes a GET request to the `.well-known/openid-configuration`
+// endpoint of BaseURL and returns the result as `OIDCWellKnown`.
+func (login *DeviceLogin) getWellKnown() (*OIDCWellKnown, error) {
+	login.BaseURL = "https://login.elixir-czech.org/oidc"
```
jbygdell, line 286:
```suggestion
```
This should not be hardcoded

main.go
```diff
@@ -61,6 +63,17 @@ func main() {
 		err = datasetsize.DatasetSize(args)
 	case "list":
 		err = list.List(args)
+	case "login":
+		deviceLogin, err := login.NewDeviceLogin(args)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to create device login: %v", err)
```
jbygdell, line 69:
```suggestion
			fmt.Fprintf(os.Stderr, "failed to contact authentication service: %v", err)
```

login/login.go
```diff
@@ -0,0 +1,390 @@
...
+
+	var url string
+	err := Args.Parse(args[1:])
+	if err != nil {
+		return DeviceLogin{}, errors.New("failed parsing arguments")
+	}
+	if len(Args.Args()) == 1 {
+		url = Args.Args()[0]
+	}
+	log.Println("url: ", url)
```
jbygdell, line 174: There are a bunch of these that looks like development debug messages rather than something the end user should see.

jbygdell (changes requested):

login/login.go
pahatz: right! testing leftovers. will remove it

README.md
```diff
@@ -204,6 +206,17 @@ Given that the instructions in the [download section](#download) have been follo
 ```
 where `<keypair_name>.sec.pem` the private key created in the [relevant section](#create-keys) and `<file_to_decrypt>` one of the files downloaded following the instructions of the [download section](#download-file).
 
+## Login
+
+You can login to download the configuration file needed for some of the the tool's operation using the login command:
+```bash
+./sda-cli login <login_target>
+```
+where `login_target` is the URL for the the RI Life Science login, or for Big Picture: https://login.bp.nbis.se/
```
dbampalikis, line 215: There is an inconsistency here and in the argHelp (`login.go` line 34) where the url is `bp.nbis.se` instead. Also, the sentence is not exactly correct, so I am changing it a bit:
```suggestion
where `login_target` is the URL can be the login endpoint for Big Picture (https://login.bp.nbis.se/), Federated EGA (https://login.test.fega.nbis.se/) or Genomic Data Infrastructure (https://login.gdi.nbis.se/)
```

login/login.go
```diff
@@ -0,0 +1,384 @@
...
+
+login:
+    logs in to the SDA using the provided login target.
+`
+
+// ArgHelp is the suffix text that will be displayed after the argument list in
+// the module help
+var ArgHelp = `
+    [login-target]
+        The login target can be one of the following: bp.nbis.se`
```
dbampalikis, line 34: Here is the inconsistency regarding the URL

login/login.go
```diff
@@ -0,0 +1,384 @@
...
+		return fmt.Errorf("failed to fetch .well-known configuration: %v", err)
+	}
+
+	login.deviceLogin, err = login.startDeviceLogin()
+	if err != nil {
+		return fmt.Errorf("failed to start device login: %v", err)
+	}
+	expires := time.Duration(login.deviceLogin.ExpiresIn * int(time.Second))
+	log.Infof("Login started (expires in %v minutes)", expires.Minutes())
+	log.Infof("Go to %v to finish logging in.", login.deviceLogin.VerificationURL)
```
dbampalikis, line 197: That's just a recommendation, in case you think it's better, we could open the default browser directly from the command line with something similar to [this](https://gist.github.com/sevkin/9798d67b2cb9d07cb05f89f14ba682f8)

dbampalikis (approved):
Good job! Some small comments, otherwise it looks good. I haven't tested it, will do once you are finished

login/login.go
pahatz: Removed, along with the debugging logs.

main.go
pahatz: added

login/login.go
pahatz: removed them.

README.md
pahatz: Good catch, updated it.

login/login.go
pahatz: fixed

login/login.go
pahatz: I added it!

pahatz:
https://github.com/NBISweden/sda-cli/pull/234

login/login.go
```diff
@@ -0,0 +1,405 @@
...
+	s3Config, err := login.GetS3Config()
+	if err != nil {
+		return err
+	}
+
+	err = ini.ReflectFrom(cfg, s3Config)
+	if err != nil {
+		return err
+	}
+	err = cfg.SaveTo(".sda-cli-session")
```
nanjiangshu, line 152: Yes, here it is a file named `.sda-cli-session` is created.

login/login.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+// Login() does a full login by fetching the remote configuration, starting the
+// login procedure, and then waiting for the user to complete login.
+func (login *DeviceLogin) Login() error {
+
+	var err error
+	login.wellKnown, err = login.getWellKnown()
+	if err != nil {
+		return fmt.Errorf("failed to fetch .well-known configuration: %v", err)
```
nanjiangshu, line 206: It's better to show the message as 

```
failed to fetch .well-known configuration, details: %v", err
```
similar as https://github.com/NBISweden/sda-cli/blob/main/download/download.go#L79 and https://github.com/NBISweden/sda-cli/blob/main/download/download.go#L90

Also, it is probably better to add a line break at the end of message, so that multiple messages will not stack in one line. 

This applies to all error messages in this script.

nanjiangshu (commented):
Great job @pahatz! I just left a few minor comments. 

There's also some unwanted loginfo which will output something like 

```
args:  [./sda-cli -config .sda-cli-session]
pos:  []
args:  [./sda-cli -config .sda-cli-session]
```
caused by changes in the script `helpers/helpers.go`. But it's weird that this file is not listed here. This refers to the print lines of https://github.com/NBISweden/sda-cli/blob/227-login-from-cli/helpers/helpers.go#L133-L138

In addition, during testing, I found that the downloaded configuration file `.sda-cli-session` is not working when I tried to access the bigpicture server. 

Should the hard-coded settings be changed, e.g. `info.InboxURI `?
```
	info.ClientID = "****"
	info.InboxURI = "s3.bp.nbis.se"
	info.OidcURI = "https://login.elixir-czech.org/oidc"   
```

README.md
```diff
@@ -204,6 +206,17 @@ Given that the instructions in the [download section](#download) have been follo
...
+## Login
+
+You can login to download the configuration file needed for some of the the tool's operation using the login command:
+```bash
+./sda-cli login <login_target>
+```
+where `login_target` is the URL can be the login endpoint for Big Picture (https://login.bp.nbis.se/), Federated EGA (https://login.test.fega.nbis.se/) or Genomic Data Infrastructure (https://login.gdi.nbis.se/)
+
+This will throw a link in the terminal for the user where they can go and log in.
+After the login is complete, a configuration file will be created in the tool's directory with the name of `sda-cli-session`
```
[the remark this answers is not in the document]
pahatz: I missed the dot, adding it

login/login.go
pahatz: I am not sure about the newline:

> error strings should not end with punctuation or newlines (ST1005)

But I see your point of it becoming a bit long with the errors following.
It seems like this is how we usually return the errors so far, error string followed by a colon and the actual error.
For example: https://github.com/NBISweden/sda-cli/blob/227d4d291b0991250236a20ff9ab9afa0919ab79/create_key/create_key.go#L54

README.md
```diff
@@ -70,14 +70,16 @@ Passing a combination of the above arguments is allowed, as well:
...
-Once your files are encrypted, they are ready to be submitted to the SDA/BP archive. The s3 storage requires users to be authenticated, therefore a configuration files needs to be downloaded before starting the uploading of the files.
+Once your files are encrypted, they are ready to be submitted to the SDA/BP archive. The s3 storage requires users to be authenticated, therefore a configuration file needs to be downloaded before starting the uploading of the files.
 
 The configuration file can be downloaded by logging in with a Life Science RI account:
 
 * For BigPicture use https://login.bp.nbis.se/
 
 The configuration file should be placed in the root folder of the repository.
 
+Alternatively, you can download the configuration file using the login command.
```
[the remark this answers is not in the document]
pahatz: added!

pahatz:
> 

You're right. I removed those log prints. 
I'll try to see how it would work on BP, I might ask for help on that.

login/login.go
nanjiangshu: It's not just the line becomes long but also messages are merged since there is no space in between. For example 

```
$ ./sda-cli login https://login.bp.nbis.se/
failed to contact authentication service: failed to get auth InfoLogin failed: failed to start device login: request failed: status code: 400
```
The message `failed to get auth Info` and `Login failed: failed to start device login: ` should be separated.

pahatz: Okay, I see now which ones are the problem. I'll try to improve it.

main.go
```diff
@@ -61,6 +63,18 @@ func main() {
 		err = datasetsize.DatasetSize(args)
 	case "list":
 		err = list.List(args)
+	case "login":
+		deviceLogin, err := login.NewDeviceLogin(args)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to contact authentication service, reason: %v\n", err)
+		}
```
jbygdell, line 70: If this fails there is no need to try to do the actual authentication request.

main.go
```diff
@@ -61,6 +63,18 @@ func main() {
...
+		err = deviceLogin.Login()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Login failed: %v\n", err)
+		} else {
+			fmt.Println("Login Successful")
+			fmt.Printf("Logged in as %v\n", deviceLogin.UserInfo.Name)
+		}
```
jbygdell, line 77: This is overly verbose.

jbygdell, line 77: Suggestion:
Move all the if-else stuff into the login.go so that it only returns an error or nil like the others.

jbygdell (changes requested):

main.go
pahatz: refactored the whole thing, so everything moved in login

pahatz: I removed the first message

login/login.go
```diff
@@ -0,0 +1,423 @@
...
+		return DeviceLogin{}, errors.New("failed parsing arguments")
+	}
+	if len(Args.Args()) == 1 {
+		url = Args.Args()[0]
+	}
+	info, err := GetAuthInfo(url)
+	fmt.Println("#####", info)
+	info.ClientID = "8b7b0168-6b16-4fd2-baec-b0a28b0d5cb0"
+	info.InboxURI = "s3.bp.nbis.se"
+	info.OidcURI = "https://login.elixir-czech.org/oidc"
```
jbygdell, line 190: Don't forget to remove stuff used for testing.

pahatz:
> Great job @pahatz! I just left a few minor comments.
> 
> There's also some unwanted loginfo which will output something like
> 
> ```
> args:  [./sda-cli -config .sda-cli-session]
> pos:  []
> args:  [./sda-cli -config .sda-cli-session]
> ```
> 
> caused by changes in the script `helpers/helpers.go`. But it's weird that this file is not listed here. This refers to the print lines of https://github.com/NBISweden/sda-cli/blob/227-login-from-cli/helpers/helpers.go#L133-L138
> 
> In addition, during testing, I found that the downloaded configuration file `.sda-cli-session` is not working when I tried to access the bigpicture server.
> 
> Should the hard-coded settings be changed, e.g. `info.InboxURI `?
> 
> ```
> 	info.ClientID = "****"
> 	info.InboxURI = "s3.bp.nbis.se"
> 	info.OidcURI = "https://login.elixir-czech.org/oidc"   
> ```

The key written in .sda-cli-session should be different for BP and EGA given how it currently works. So this issue 
(if there is one) should be handled on the sda-auth part of the login. sda-auth should provide the right key/configs depending on the deployment it's running on.

main.go
pahatz: This was also changed as I refactored to move that block out of main.go

Issue #236, opened 2023-07-06
Use .sda-cli-session as a fallback if no config is provided in upload and list options


pahatz:
Currently the login option of the cli creates the .sda-cli-session file which is identical to the config file.
The cli should try on `list` and `upload` - that require a config - fall back to look for the .sda-cli-session file to use if no config is provided.

PR #237, opened 2023-07-06
Fallback to .sda-cli-session if no config is provided

pahatz:
Currently the login option of the cli creates the .sda-cli-session file which is identical to the config file.
The cli should try on list and upload - that require a config - fall back to look for the .sda-cli-session file to use if no config is provided.

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -275,6 +275,19 @@ func LoadConfigFile(path string) (*Config, error) {
 	return config, nil
 }
 
+// GetAuth calls LoadConfig if we have a config file, otherwise try to load .sda-cli-session
+func GetAuth(path string) (*Config, error) {
+
+	if path != "" {
+		return LoadConfigFile(path)
+	}
+	if FileExists(".sda-cli-session") {
+		return LoadConfigFile(".sda-cli-session")
+	}
+
+	return nil, errors.New("failed to read the configuration file")
+}
+
 // CheckTokenExpiration is used to determine whether the token is expiring in less than a day
 func CheckTokenExpiration(accessToken string) (bool, error) {
 
--- a/list/list_test.go
+++ b/list/list_test.go
@@ -39,7 +39,7 @@ func (suite *TestSuite) TestNoConfig() {
 	os.Args = []string{"list", "-config", ""}
 
 	err := List(os.Args)
-	assert.EqualError(suite.T(), err, "failed to find an s3 configuration file for listing data")
+	assert.EqualError(suite.T(), err, "failed to load config file, reason: failed to read the configuration file")
 }
 
 func (suite *TestSuite) TestTooManyArgs() {
--- a/main.go
+++ b/main.go
@@ -113,6 +113,12 @@ func ParseArgs() (string, []string) {
 		Help(subcommand)
 	}
 
+	// list command can have no arguments since it can use the config from login
+	// so we immediately return in that case
+	if command == "list" {
+		return command, os.Args
+	}
+
 	// If no arguments are provided to the subcommand, it's not gonna be valid,
 	// so we print the subcommand help
 	if len(os.Args) == 1 {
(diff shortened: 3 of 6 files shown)


pahatz:
closes #236

nanjiangshu (changes requested):
Looks good @pahatz!

However, shouldn't this branch be based on `227-login-from-cli`? Otherwise, it is not clear to see what additional changes are compared to the PR 227. 

Also,  argument parsing for the `list` sub-command (in the file `list/list.go` needs to be changed as well, otherwise when running the command 
```
./sda-cli list
```
it will just show the help message instead of listing the content of files in the s3 bucket.

pahatz:
> Looks good @pahatz!
> 
> However, shouldn't this branch be based on `227-login-from-cli`? Otherwise, it is not clear to see what additional changes are compared to the PR 227.
> 
> Also, argument parsing for the `list` sub-command (in the file `list/list.go` needs to be changed as well, otherwise when running the command
> 
> ```
> ./sda-cli list
> ```
> 
> it will just show the help message instead of listing the content of files in the s3 bucket.

Yes, as we discussed this became a duplicate as I changed the tickets and it was no longer needed. I'll close it as soon as we merge  PR 227.

pahatz:
> Looks good @pahatz!
> 
> However, shouldn't this branch be based on `227-login-from-cli`? Otherwise, it is not clear to see what additional changes are compared to the PR 227.
> 
> Also, argument parsing for the `list` sub-command (in the file `list/list.go` needs to be changed as well, otherwise when running the command
> 
> ```
> ./sda-cli list
> ```
> 
> it will just show the help message instead of listing the content of files in the s3 bucket.

Good catch on the list command. It should work now fine with the latest push.

helpers/helpers.go
```diff
@@ -275,6 +275,19 @@ func LoadConfigFile(path string) (*Config, error) {
...
+func GetAuth(path string) (*Config, error) {
+
+	if path != "" {
+		return LoadConfigFile(path)
+	}
+	if FileExists(".sda-cli-session") {
+		return LoadConfigFile(".sda-cli-session")
+	}
+
+	return nil, errors.New("Failed to read the configuration file")
```
aaperis, line 288:
```suggestion
	return nil, errors.New("failed to read the configuration file")
```
since this is a propagated error message

list/list_test.go
```diff
@@ -39,7 +39,7 @@ func (suite *TestSuite) TestNoConfig() {
 	os.Args = []string{"list", "-config", ""}
 
 	err := List(os.Args)
-	assert.EqualError(suite.T(), err, "failed to find an s3 configuration file for listing data")
+	assert.EqualError(suite.T(), err, "failed to load config file, reason: Failed to read the configuration file")
```
aaperis, line 42:
```suggestion
	assert.EqualError(suite.T(), err, "failed to load config file, reason: failed to read the configuration file")
```

main.go
```diff
@@ -113,6 +113,12 @@ func ParseArgs() (string, []string) {
 		Help(subcommand)
 	}
 
+	// list command can have no arguments since it can use the config from login
+	// so we immediately return in that case
+	if command == "list" {
+		return command, os.Args
+	}
```
aaperis, line 120:
```suggestion
```

otherwise this changes the general behavior pattern that subcommand without arguments prints the subcommand's usage. See comment in code  right below.

aaperis (changes requested):
Changes seem reasonable apart from the code that circumvents the behavior of the `list` module :-)

main.go
pahatz: Yes but now list command can work without arguments since we it can work with the .sda-cli-session as a config from a previous login.

pahatz: so I should update the usage and readme I guess.

aaperis: oh right, then yes :-)

pahatz: added

aaperis (approved):

Issue #232, opened 2023-06-21
Add version command


dbampalikis:
As an sda-cli user
I want to be able to check the sda-cli version
In order to make sure I am up-to-date (or something)

Add a subcommand in sda-cli that returns the version of the cli (the release version)
 Is it hardcoded or is there any better solution to tackle that?

PR #239, opened 2023-07-13
#232: add version command

pahatz:
Added the command version that returns the current version of the cli.
The default version is defined in main and is equal to `development`.
An ldflag has been added to the golreleaser config that sets the GORELEASER_CURRENT_TAG to that variable in main.
It then gets picked up in the version implementation and is returned to the user.
Some modifications to the parsing of the arguments were needed because of the special case of version command having only two arguments.
closes: #232

### DIFF
--- a/main.go
+++ b/main.go
@@ -81,14 +81,16 @@ func main() {
 func ParseArgs() (string, []string) {
 
 	// Print usage if no arguments are provided
-	if os.Args[1] == "version" || os.Args[1] == "-v" || os.Args[1] == "--ver" {
+	if len(os.Args) < 2 {
+		Help("help")
+	}
+
+	if os.Args[1] == "version" || os.Args[1] == "-v" || os.Args[1] == "--version" {
 		if len(os.Args) != 2 {
 			Help("version")
 		}
 
 		return "version", os.Args
-	} else if len(os.Args) < 2 {
-		Help("help")
 	}
 
 	// Extract `command` from arg 1, then remove it from the flag list.
--- a/main.go
+++ b/main.go
@@ -13,8 +13,11 @@ import (
 	"github.com/NBISweden/sda-cli/helpers"
 	"github.com/NBISweden/sda-cli/list"
 	"github.com/NBISweden/sda-cli/upload"
+	"github.com/NBISweden/sda-cli/version"
 	log "github.com/sirupsen/logrus"
 )
+
+var Version = "development"
 
 var Usage = `USAGE: %s <command> [command-args]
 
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -75,7 +81,13 @@ func main() {
...
 
 	// Print usage if no arguments are provided
-	if len(os.Args) < 2 {
+	if os.Args[1] == "version" || os.Args[1] == "-v" || os.Args[1] == "--ver" {
+		if len(os.Args) != 2 {
+			Help("version")
+		}
+
+		return "version", os.Args
+	} else if len(os.Args) < 2 {
```
nanjiangshu, line 90: Should place this if condition at the top, otherwise the command will return `index out of rage` error when no argument is provided.

main.go
```diff
@@ -13,9 +13,12 @@ import (
 	"github.com/NBISweden/sda-cli/helpers"
 	"github.com/NBISweden/sda-cli/list"
 	"github.com/NBISweden/sda-cli/upload"
+	"github.com/NBISweden/sda-cli/version"
 	log "github.com/sirupsen/logrus"
 )
 
+var Version = "development"
```
nanjiangshu, line 20: Should the value of `version` be configurable?

nanjiangshu (changes requested):
Great that we now have a flag to show the version of the command. I left a few minor comments.

main.go
pahatz: yes, the goreleaser will be able to pick up the git tag we create every time we release a new version. then it overwrites this value with the GORELEASER_CURRENT_TAG.
This is like a default value in case someone creates something like an unofficial build.

main.go
pahatz: good catch!

nanjiangshu (approved):
LGTM

main.go
```diff
@@ -13,9 +13,12 @@ import (
 	"github.com/NBISweden/sda-cli/helpers"
```
nanjiangshu, line 13: In many `import` cases, packages such as `helpers` that are included in the repository are imported as remote repos. I don't quite understand the benefit of it. May someone explain it @pahatz @norling @dbampalikis ?

nanjiangshu: Sorry, I didn't realized that the project name is defined as `github.com/NBISweden/sda-cli`  in https://github.com/NBISweden/sda-cli/blob/main/go.mod#L1. So this is a local importing.

main.go
```diff
@@ -79,6 +85,14 @@ func ParseArgs() (string, []string) {
 		Help("help")
 	}
 
+	if os.Args[1] == "version" || os.Args[1] == "-v" || os.Args[1] == "--ver" {
```
aaperis, line 88: is `--ver` necessary? It doesn't seem very common.

aaperis (approved):
I wonder if the `--ver` flag is necessary, otherwise it looks good to me :)

main.go
pahatz: Could go without it. A third opinion to rule it?

jbygdell: Only `version` or `--version` should be necessary.

jbygdell (approved):
Approved with comment

main.go
nanjiangshu: Maybe just remove it.

Issue #240, opened 2023-07-14
Use the public key provided from OIDC login in encrypt command


pahatz:
This ticket should sort out the changes needed on the encrypt command to take into account the login session.
At this point we should have an endpoint in auth that returns:
`{
   "client_id" : "XC56EL11xx",
   "inbox_uri" : "s3.example.com",
   "oidc_uri" : "http://auth:8080",
   "public_key" : "6246gr445fd9e3944642624646a55e61e3801fc845f63781a8j4k4j36k4jh624a"
}`
The cli can already read the auth endpoint and login the user using OIDC, writes then the config in the .sda-cli-session file.

Now the encrypt should try after reading all the keys that the user gives as arguments to read the session file and then find that public key entry.
Then call a func in helpers that should be able to use that pub key string to create a pub key file with a name like "key-from-oidc.pub.pem" in cli root dir.
then use that key instead of failing.

* It should also be used in upload with encrypt

PR #241, opened 2023-07-14
#240: Use login pub key

pahatz:
Encrypt reads all the keys that the key arguments from the user and then reads the session file to get the public key entry.
Then calls a func in helpers that creates a file "key-from-oidc.pub.pem" in cli root dir that contains that public key from login.
Last, it uses that key to encrypt instead of failing.
closes: #240

### DIFF
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -75,9 +75,14 @@ func Encrypt(args []string) error {
 		return err
 	}
 
-	// Exit if public key is not provided
+	// no key provided, check for one in the session file
 	if len(publicKeyFileList) == 0 {
-		return fmt.Errorf("public key not provided")
+
+		sesKey, err := helpers.GetPublicKey()
+		if err != nil {
+			return fmt.Errorf("public key not provided, details: %v", err)
+		}
+		publicKeyFileList = append(publicKeyFileList, sesKey)
 	}
 
 	// Each filename is first read into a helper struct (sliced for combatibility with checkFiles)
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/golang-jwt/jwt"
 	"github.com/manifoldco/promptui"
+	"github.com/neicnordic/crypt4gh/keys"
 	log "github.com/sirupsen/logrus"
 	"github.com/vbauerster/mpb/v8"
 	"golang.org/x/exp/slices"
@@ -227,6 +228,7 @@ type Config struct {
 	UseHTTPS             bool   `ini:"use_https"`
 	SocketTimeout        int    `ini:"socket_timeout"`
 	HumanReadableSizes   bool   `ini:"human_readable_sizes"`
+	PublicKey            string `ini:"public_key"`
 }
 
 // LoadConfigFile loads ini configuration file to the Config struct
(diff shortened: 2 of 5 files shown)


encrypt/encrypt.go
```diff
@@ -137,6 +143,7 @@ func Encrypt(args []string) error {
 	// Read the public key(s) to be used for encryption. The matching private
 	// key will be able to decrypt the file.
 	pubKeyList, err := createPubKeyList(publicKeyFileList, c4ghKeySpecs)
+	// fmt.Println("pubKeyList: ", pubKeyList)
```
nanjiangshu, line 146: This line could be removed

nanjiangshu (commented):
Looks great @pahatz . I changed the base branch to `227-login-from-cli-2` for easy reviewing. Apartment from that, I left a few minor comments.

encrypt/encrypt.go
```diff
@@ -75,9 +75,15 @@ func Encrypt(args []string) error {
...
-	// Exit if public key is not provided
+	// no key provided, check for one in the session file
 	if len(publicKeyFileList) == 0 {
-		return fmt.Errorf("public key not provided")
+
+		sesKey, err := helpers.GetPublicKey()
+		if err != nil {
+			return fmt.Errorf("public key not provided")
+		}
+		// fmt.Println("sesKey: ", sesKey)
```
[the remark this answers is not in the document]
pahatz: ah right, always some leftover logs :D

encrypt/encrypt.go
```diff
@@ -75,9 +75,15 @@ func Encrypt(args []string) error {
...
 	}
 
-	// Exit if public key is not provided
+	// no key provided, check for one in the session file
 	if len(publicKeyFileList) == 0 {
-		return fmt.Errorf("public key not provided")
+
+		sesKey, err := helpers.GetPublicKey()
+		if err != nil {
+			return fmt.Errorf("public key not provided")
```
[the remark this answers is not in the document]
pahatz: added

nanjiangshu (approved):

helpers/helpers.go
```diff
@@ -288,6 +290,50 @@ func GetAuth(path string) (*Config, error) {
 	return nil, errors.New("failed to read the configuration file")
 }
 
+func GetPublicKey() (string, error) {
```
aaperis, line 293: It would be useful to add specific tests in the `helpers_test.go` suite for all new functions.

aaperis:
It looks good! Since you're at it, it would be nice to add to the integration test script a happy path where `.sda-cli-session` exists (or split this to a new PR).

helpers/helpers.go
aaperis: I meant `GetPublicKey() (string, error)` etc, not the ones imported :-)

encrypt/encrypt.go
```diff
@@ -75,9 +75,14 @@ func Encrypt(args []string) error {
...
 	}
 
-	// Exit if public key is not provided
+	// no key provided, check for one in the session file
 	if len(publicKeyFileList) == 0 {
-		return fmt.Errorf("public key not provided")
+
+		sesKey, err := helpers.GetPublicKey()
+		if err != nil {
+			return fmt.Errorf("public key not provided, details: %v", err)
```
aaperis, line 83:
```suggestion
			return fmt.Errorf("public key not provided or %v", err)
```
so that the error msg applies covers also the case when a user intends to input a public key in the command line.

helpers/helpers.go
pahatz: added some

pahatz:
> It looks good! Since you're at it, it would be nice to add to the integration test script a happy path where `.sda-cli-session` exists (or split this to a new PR).

I added some tests on the func level, I think they would be testing pretty much the same thing

helpers/helpers.go
```diff
@@ -288,6 +290,57 @@ func GetAuth(path string) (*Config, error) {
...
+	// Check if the ".sda-cli-session" file exists
+	if !FileExists(".sda-cli-session") {
+		return "", errors.New("configuration file (.sda-cli-session) not found")
+	}
+
+	if FileExists(".sda-cli-session") {
+		file, err := os.Open(".sda-cli-session")
+		if err != nil {
+			fmt.Println("######file:", file)
```
aaperis, line 302:
```suggestion
			fmt.Println("could not read file:", file)
```

helpers/helpers_test.go
```diff
@@ -305,3 +305,71 @@ func (suite *HelperTests) TestTokenExpiration() {
...
+	defer os.Remove(configPath.Name())
+
+	err = os.WriteFile(configPath.Name(), []byte(confFile), 0600)
+	if err != nil {
+		log.Printf("failed to write temp config file, %v", err)
+	}
+
+	_, err = GetPublicKey()
+	assert.NoError(suite.T(), err)
+}
```
aaperis, line 375:
```suggestion
	if assert.FileExists(suite.T(), "key-from-oidc.pub.pem") {
		os.Remove("key-from-oidc.pub.pem")
	}
}
```
for cleaning up after test

helpers/helpers.go
pahatz: changed

aaperis (approved):
Looks good!

PR #255, opened 2023-08-30
Cleanup after crypt4gh update

aaperis:
After the fix in https://github.com/neicnordic/crypt4gh/pull/85 there is no need anymore for the `sda-cli` to handle panics from the crypt4gh library, so this logic is now removed. Since `crypt4gh` now returns an error msgs instead, some tests for the `sda-cli` need to be updated which is also done in this PR.

Closes PR #254


kusalananda (approved):

encrypt/encrypt.go
```diff
@@ -326,36 +326,21 @@ func calculateHashes(fileSet helpers.EncryptionFileSet) (*hashSet, error) {
 }
 
 // Reads a public key from a file using the crypt4gh keys module
-func readPublicKeyFile(filename string) (key *[32]byte, err error) {
```
pahatz, line 329: Doesn't it make more sense to keep this func name as is?
We still call keys.ReadPublicKey that does the actual pub key reading.

pahatz: but that call is still inside here... not sure really :)

pahatz (approved):
Looks good!

encrypt/encrypt.go
aaperis: hmm good point. I originally kept that name for parity since `decrypt.go` has a respective `readPrivateKey` function.  But they should both be renamed since they both read from file whereas the respective functions from the `keys` library of `crypt4gh` read from stream. So, I renamed them.

PR #256, opened 2023-09-01
Credentials are access key, secret key, access token

pontus:
Not access key, access key, access token.


aaperis:
I am guessing that configuration was hardcoded in this way so that `sda-cli` works with  `s3-proxy` when passing standard  `s3cmd` config files. If these lines change, we should check whether parts of the  documentation need to change as well.

pontus:
Possibly, but the configuration as sent by sda-auth does it the right way around (for s3proxy) with secret_key and access_key set to the same thing already, so I can't see why this would be needed.

dbampalikis (approved):

aaperis:
Ok then I do not see any problem with the change.

aaperis (approved):

PR #274, opened 2023-10-24
Test with latest s3inbox

### LINKED ISSUE #273: Use latest s3inbox image for testing
The `s3inbox` (s3-proxy) image used for integration testing is quite old so we need to keep it updated in order to ensure compatibility with the `sda-cli` as more feature/changes happen.

aaperis:
This PR updates the s3inbox image in the integration tests to the latest available in the merged repository (v0.0.130) and updates the configuration accordingly. The token signing script has been replaced with the one taken from that repository so that the token can be properly verified by the s3inbox.

The latest image tag for `sensitive-data-archive` images is used programmatically during integration testing (thanks to @jbygdell) .
Also fixed a silent failure in the integration tests (thanks @kusalananda) and added `exit 1` status for the app when it exits with error.

Closes #273


testing/docker-compose.yml
```diff
@@ -48,7 +48,8 @@ services:
       timeout: 20s
       retries: 3
   s3_proxy:
-    image: ghcr.io/neicnordic/sda-s3proxy:latest
+    image: ghcr.io/neicnordic/sensitive-data-archive:v0.0.130
```
jbygdell, line 51:
```suggestion
    image: ghcr.io/neicnordic/sensitive-data-archive:${TAG}
```

If you consistently want to test against the latest released version, set the tag as a variable like above.
Then you can start the compose file with the tag set externally:

```cmd
TAG="v0.0.139" docker compose up -d
```

And you can make use of the following command to dynamically get the latest released tag from the `sensitive-data-archive` repo:

```cmd
curl -s https://api.github.com/repos/neicnordic/sensitive-data-archive/tags | jq -r '.[0].name'
```

Putting it all together yields:

```cmd
TAG=$(curl -s https://api.github.com/repos/neicnordic/sensitive-data-archive/tags | jq -r '.[0].name') docker compose up -d
```

jbygdell (approved):
LGTM

Although take a look at my comment so we don't have to update this compose file every week in order to keep it up to date.

.github/integration/tests/tests.sh
```diff
@@ -92,7 +95,7 @@ cp data_files_enc/data_file.c4gh data_files_enc/dir1/dir2/data_file2.c4gh
 # Check that files were uploaded with the local path prefix `data_files_enc` stripped from the target path
 for k in dir1/data_file.c4gh dir1/dir2/data_file.c4gh dir1/dir2/data_file2.c4gh data_file3.c4gh
 do
-    check_uploaded_file test/dummy/$k $k
+    check_uploaded_file test/"$user"/$k $k
```
kusalananda, line 98:
```suggestion
    check_uploaded_file "test/$user/$k" "$k"
```

.github/integration/tests/tests.sh
```diff
@@ -47,7 +50,7 @@ check_encypted_file $files
 
 # Upload a specific file and check it
 ./sda-cli upload -config testing/s3cmd.conf data_file.c4gh
-check_uploaded_file test/dummy/data_file.c4gh data_file.c4gh
+check_uploaded_file test/"$user"/data_file.c4gh data_file.c4gh
 
 output=$(./sda-cli list -config testing/s3cmd.conf | grep -q "data_file.c4gh")
```
kusalananda, line 55: Note that `"$output"` will _always_ be empty since `grep -q` never outputs anything.

What I think is meant is this:
```
if ./sda-cli list -config testing/s3cmd.conf | grep -q 'data_file\.c4gh'
then
    echo 'Listed file from S3 backend'
else
    echo 'Failed to list file from S3 backend'
    exit 1
fi
```

.github/integration/tests/tests.sh
```diff
@@ -287,7 +290,7 @@ fi
 # Try downloading from private bucket
 printf "%s" "Attempting to download from S3 bucket with ACL=private..."
 errorMessage="reason: request failed with \`403 Forbidden\`, details: {Code:AllAccessDisabled"
```
kusalananda, line 292: Put the message in single quotes to avoid having to escape the backticks:
```suggestion
errorMessage='reason: request failed with `403 Forbidden`, details: {Code:AllAccessDisabled'
```

.github/integration/tests/tests.sh
```diff
@@ -135,10 +138,10 @@ check_encypted_file "data_files_unenc/data_file.c4gh" "data_files_unenc/dir1/dat
...
 do
-    check_uploaded_file test/dummy/$uploadDir"/"$k $k
+    check_uploaded_file test/"$user"/$uploadDir"/"$k $k
 done
 
-out_ls=$(s3cmd -c testing/directS3 ls -r s3://test/dummy/testEncryptUpload/data_files_unenc/)
+out_ls=$(s3cmd -c testing/directS3 ls -r s3://test/"$user"/testEncryptUpload/data_files_unenc/)
 if ( ! echo "$out_ls" | grep -v -q "c4gh" ); then
     echo "No unencrypted files were uploaded during encrypt+upload"
 else
```
kusalananda, line 147: No reason to use an intermediate variable here:
```suggestion
if ! s3cmd -c testing/directS3 ls -r "s3://test/$user/testEncryptUpload/data_files_unenc/" | grep -v -q 'c4gh'
then
    echo 'No unencrypted files were uploaded during encrypt+upload'
else
```

.github/integration/tests/tests.sh
```diff
@@ -108,7 +111,7 @@ uploadDir="testfolder"
 # target path and into the specified upload folder
 for k in dir1/data_file.c4gh dir1/dir2/data_file.c4gh dir1/dir2/data_file2.c4gh data_file3.c4gh
 do
-    check_uploaded_file test/dummy/$uploadDir"/"$k $k
+    check_uploaded_file test/"$user"/$uploadDir"/"$k $k
```
kusalananda, line 114:
```suggestion
    check_uploaded_file "test/$user/$uploadDir/$k" "$k"
```

kusalananda (changes requested):
Some quoting fixes, and also other comments.  The most interesting find here is a `grep -q` that the code seems to expect to output something (it never does), but that wasn't even part of the code touched by this PR, just adjacent to it.

.github/integration/tests/tests.sh
aaperis: Indeed, good catch!

.github/integration/tests/tests.sh
aaperis: this made shellcheck complain so I left it as it was

jbygdell (approved):

kusalananda (approved):
All good

PR #279, opened 2023-11-06
Fix/225 help exit code

### LINKED ISSUE #225: `sda-cli` should not exit with non-zero status when using `help` subcommand
Asking for help is not an error, so `sda-cli help ...` should not exit with a non-zero exit status (which it currently does).  Only when the utility terminates due to an actual error should the exit stats be non-zero.

kusalananda:
This closes #225 

The PR moves the actual `os.Exit()` call out of the `Help()` function, which allows the caller to exit with either a zero or non-zero exit status, depending on the circumstances.

### DIFF
--- a/main.go
+++ b/main.go
@@ -89,11 +89,13 @@ func ParseArgs() (string, []string) {
 	// Print usage if no arguments are provided
 	if len(os.Args) < 2 {
 		Help("help")
+		os.Exit(1)
 	}
 
 	if os.Args[1] == "version" || os.Args[1] == "-v" || os.Args[1] == "--version" {
 		if len(os.Args) != 2 {
 			Help("version")
+			os.Exit(1)
 		}
 
 		return "version", os.Args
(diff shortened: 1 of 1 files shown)


nanjiangshu (approved):
Looks good! I actually think `./sda-cli help` should also return a zero since there's no error. But I approve it anyway since it meets the description of the issue.

kusalananda:
> Looks good! I actually think `./sda-cli help` should also return a zero since there's no error. But I approve it anyway since it meets the description of the issue.

I thought about that, but since `help` is not documented as a subcommand, and since there is an explicit mention in the code about `list` not taking any arguments, but no similar special treatment of `help`, I made the decision to treat a lone `help` as an error.

aaperis (approved):
Looks good :-)

kostas-kou (approved):

pahatz (approved):
Looks great! It seems to do the trick.

PR #285, opened 2023-11-14
Encrypt fetches pub key from auth

### LINKED ISSUE #264: Encrypt fetches pub key from auth
Updates needed:
 - [x] Endpoint implemented in neicnordic/sensitive-data-archive#200
 - [x] `sda-cli` to call it and initate the process as above
 This was already correctly implemented
 - [ ] there's a change in the config's fields for access_key to replace the `@` with `_`
 - [x] encrypt needs to look for the pub key directly in auth info endpoint as well. the current implementation only checks if it was put it .sda-cli-session config from a previous call on ./sda-cli login

pahatz:
closes #264 
Change encrypt module to fetch the public key from auth's endpoint info. Encrypt will now try to find a public key in the following order from:
- provided as an argument
- if `target` was provided, fetch it
- lookup in the .sda-cli-session from a previous login

### DIFF
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -28,11 +28,6 @@ import (
 var Usage = `
 USAGE: %s encrypt -key <public-key-file> (-target <target>) (-outdir <dir>) (-continue=true) [file(s)]
 
-The target can be one of the following:
-	bp.nbis.se
-	fega.nbis.se
-	gdc.nbis.se
-
 encrypt:
     Encrypts files according to the crypt4gh standard used in the
     Sensitive Data Archive (SDA).  Each given file will be encrypted
@@ -82,36 +77,36 @@ func Encrypt(args []string) error {
 		return err
 	}
 
-	var sesKey string
-	if len(publicKeyFileList) == 0 {
-		// check for public key in .sda-cli-session file from login
-		sesKey, err = helpers.GetPublicKeyFromSession()
-		if err != nil {
-			log.Println("could not read key from previous login,", err)
-		}
+	if publicKeyFileList != nil && *target != "" {
+		return errors.New("only one of -key or -target can be used")
 	}
-	// key from session file found
-	if len(publicKeyFileList) == 0 && sesKey != "" {
-		publicKeyFileList = append(publicKeyFileList, sesKey)
-	}
-	if len(publicKeyFileList) == 0 && sesKey == "" && *target != "" {
+
+	if *target != "" {
 		// fetch info endpoint values
+		log.Println("fetching public key")
 		info, err := login.GetAuthInfo(*target)
 		if err != nil {
 			return err
 		}
 		// create pub file
-		pubFile, err := helpers.CreatePubFile(info.PublicKey, "crypt4gh_key.pub")
+		pubKeyFile, err := helpers.CreatePubFile(info.PublicKey, "crypt4gh_key.pub")
 		if err != nil {
 			return err
 		}
-		log.Println("fetching public key")
 		// no key provided, no key in session file, target provided
-		publicKeyFileList = append(publicKeyFileList, pubFile)
+		publicKeyFileList = append(publicKeyFileList, pubKeyFile)
 	}
 	// no key provided, no key in session file, no target provided
-	if len(publicKeyFileList) == 0 && sesKey == "" && *target == "" {
-		return errors.New("no public key could be obtained")
+	if publicKeyFileList == nil && *target == "" {
+		// check for public key in .sda-cli-session file from login
+		pubKey, err := helpers.GetPublicKeyFromSession()
+		if err != nil {
+			return err
+		}
+		// key from session file found
+		if len(publicKeyFileList) == 0 && pubKey != "" {
+			publicKeyFileList = append(publicKeyFileList, pubKey)
+		}
 	}
 
 	// Each filename is first read into a helper struct (sliced for combatibility with checkFiles)
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -15,7 +15,7 @@ import (
 	"strings"
 
 	"github.com/NBISweden/sda-cli/helpers"
-
+	"github.com/NBISweden/sda-cli/login"
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/crypt4gh/streaming"
 	log "github.com/sirupsen/logrus"
@@ -26,7 +26,12 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help encrypt` command
 var Usage = `
-USAGE: %s encrypt -key <public-key-file> (-outdir <dir>) (-continue=true) [file(s)]
+USAGE: %s encrypt -key <public-key-file> (-target <target>) (-outdir <dir>) (-continue=true) [file(s)]
+
+The target can be one of the following:
+	bp.nbis.se
+	fega.nbis.se
+	gdc.nbis.se
 
 encrypt:
     Encrypts files according to the crypt4gh standard used in the
@@ -54,6 +59,8 @@ var outDir = Args.String("outdir", "",
 
 var continueEncrypt = Args.Bool("continue", false, "Do not exit on file errors but skip and continue.")
 
+var target = Args.String("target", "", "Client target for public key.")
+
 var publicKeyFileList []string
 
 func init() {
@@ -75,14 +82,36 @@ func Encrypt(args []string) error {
 		return err
 	}
 
-	// no key provided, check for one in the session file
+	var sesKey string
 	if len(publicKeyFileList) == 0 {
-
-		sesKey, err := helpers.GetPublicKey()
+		// check for public key in .sda-cli-session file from login
+		sesKey, err = helpers.GetPublicKeyFromSession()
 		if err != nil {
-			return fmt.Errorf("public key not provided or %v", err)
+			log.Println("could not read key from previous login,", err)
 		}
+	}
+	// key from session file found
+	if len(publicKeyFileList) == 0 && sesKey != "" {
 		publicKeyFileList = append(publicKeyFileList, sesKey)
+	}
+	if len(publicKeyFileList) == 0 && sesKey == "" && *target != "" {
+		// fetch info endpoint values
+		info, err := login.GetAuthInfo(*target)
+		if err != nil {
+			return err
+		}
+		// create pub file
+		pubFile, err := helpers.CreatePubFile(info.PublicKey, "crypt4gh_key.pub")
+		if err != nil {
+			return err
+		}
+		log.Println("fetching public key")
+		// no key provided, no key in session file, target provided
+		publicKeyFileList = append(publicKeyFileList, pubFile)
+	}
+	// no key provided, no key in session file, no target provided
+	if len(publicKeyFileList) == 0 && sesKey == "" && *target == "" {
+		return errors.New("no public key could be obtained")
 	}
 
 	// Each filename is first read into a helper struct (sliced for combatibility with checkFiles)
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -299,11 +299,9 @@ func GetPublicKeyFromSession() (string, error) {
 		return "", errors.New("configuration file (.sda-cli-session) not found")
 	}
 
-	if FileExists(".sda-cli-session") {
-		file, err := os.Open(".sda-cli-session")
-		if err != nil {
-			fmt.Println("could not read file:", file)
-		}
+	_, err := os.Open(".sda-cli-session")
+	if err != nil {
+		return "", err
 	}
 
 	// Load the configuration file
@@ -319,7 +317,7 @@ func GetPublicKeyFromSession() (string, error) {
 
 	pubFile, err := CreatePubFile(config.PublicKey, "key-from-oidc.pub.pem")
 	if err != nil {
-		return "", fmt.Errorf("failed to create public key file: %w", err)
+		return "", err
 	}
 
 	return pubFile, nil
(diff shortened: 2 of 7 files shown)


encrypt/encrypt.go
```diff
@@ -75,15 +82,37 @@ func Encrypt(args []string) error {
...
+			return err
+		}
+		log.Println("fetching public key")
+		// no key provided, no key in session file, target provided
+		publicKeyFileList = append(publicKeyFileList, pubFile)
+	}
+	// no key provided, no key in session file, no target provided
+	if len(publicKeyFileList) == 0 && sesKey == "" && *target == "" {
+		return errors.New("no public key could be obtained")
+	}
```
jbygdell, line 115: This entire block makes no logical sense.
 `GetPublicKeyFromSession` will return the name of the public key or an error.
That means that this block can be simplified to two if blocks:
```
if publicKeyFileList == nil && *target == "" {
...
```
and
```
if publicKeyFileList == nil && *target != "" {
...
```
or as a switch statement:
```
if publicKeyFileList == nil {
	switch len(*target) {
	case 0:
...
```

jbygdell (changes requested):

aaperis:
After discussions, the suggestion is that the `-target` flag should be mandatory so as to avoid reading info from the `.session` file that is for the wrong project.

jbygdell:
> After discussions, the suggestion is that the `-target` flag should be mandatory so as to avoid reading info from the `.session` file that is for the wrong project.

That is not a good idea.
First of all, you can not prevent people from doing stupid things.
Second, it will make `sda-cli` unusable in airgaped/offline environments.

And if you look in the code the session file is stored in the current working directory, so the chance of a user having the "wrong" session file is slim, unless the user executes all commands from the exact same location.

The proper way to handle this would be to first sketch up the different scenarios for our 3 cases:
* specific public key - simple: if defined, use it.
* session file - last resort if neither `-key` nor `-target` is supplied
* `target` option - simple: if defined, use it.

Only one of `-key` or `-target` should be allowed.

pahatz:
> > After discussions, the suggestion is that the `-target` flag should be mandatory so as to avoid reading info from the `.session` file that is for the wrong project.
> 
> That is not a good idea. First of all, you can not prevent people from doing stupid things. Second, it will make `sda-cli` unusable in airgaped/offline environments.
> 
> And if you look in the code the session file is stored in the current working directory, so the chance of a user having the "wrong" session file is slim, unless the user executes all commands from the exact same location.
> 
> The proper way to handle this would be to first sketch up the different scenarios for our 3 cases:
> 
>     * specific public key - simple: if defined, use it.
> 
>     * session file - last resort if neither `-key` nor `-target` is supplied
> 
>     * `target` option - simple: if defined, use it.
> 
> 
> Only one of `-key` or `-target` should be allowed.

I changed the order of the checks to respect this order now.

encrypt/encrypt.go
```diff
@@ -26,7 +26,12 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help encrypt` command
 var Usage = `
-USAGE: %s encrypt -key <public-key-file> (-outdir <dir>) (-continue=true) [file(s)]
+USAGE: %s encrypt -key <public-key-file> (-target <target>) (-outdir <dir>) (-continue=true) [file(s)]
+
+The target can be one of the following:
+	bp.nbis.se
+	fega.nbis.se
+	gdc.nbis.se
```
jbygdell, line 34:
```suggestion
```

remove all references to specific sites

encrypt/encrypt.go
```diff
@@ -75,14 +82,41 @@ func Encrypt(args []string) error {
...
-	// no key provided, check for one in the session file
-	if len(publicKeyFileList) == 0 {
+	var pubKey string
 
-		sesKey, err := helpers.GetPublicKey()
+	if publicKeyFileList != nil && *target != "" {
+		return errors.New("only one of -key or -target can be used")
+	}
+
+	if publicKeyFileList == nil && *target != "" {
```
jbygdell, line 91:
```suggestion
	if *target != "" {
```

Since only one of them can be set at this point we only need to check the one we are interested in.

encrypt/encrypt.go
```diff
@@ -75,14 +82,41 @@ func Encrypt(args []string) error {
...
+		if err != nil {
+			log.Println("could not read key from previous login,", err)
+		}
+		// key from session file found
+		if len(publicKeyFileList) == 0 && pubKey != "" {
+			publicKeyFileList = append(publicKeyFileList, pubKey)
+		}
+	}
+	if publicKeyFileList == nil {
+		return errors.New("no public key could be obtained")
```
jbygdell, line 119:
```suggestion
```
either `publicKeyFileList` has entries or we have returned an error by now.

encrypt/encrypt.go
```diff
@@ -75,14 +82,41 @@ func Encrypt(args []string) error {
...
+		log.Println("fetching public key")
+		// no key provided, no key in session file, target provided
+		publicKeyFileList = append(publicKeyFileList, pubKeyFile)
+	}
+	// no key provided, no key in session file, no target provided
+	if publicKeyFileList == nil && *target == "" {
+		// check for public key in .sda-cli-session file from login
+		pubKey, err = helpers.GetPublicKeyFromSession()
+		if err != nil {
+			log.Println("could not read key from previous login,", err)
```
jbygdell, line 111:
```suggestion
			return err
```

encrypt/encrypt.go
```diff
@@ -75,14 +82,41 @@ func Encrypt(args []string) error {
...
+		}
+		// key from session file found
+		if len(publicKeyFileList) == 0 && pubKey != "" {
+			publicKeyFileList = append(publicKeyFileList, pubKey)
+		}
```
jbygdell, line 116:
```suggestion
		publicKeyFileList = append(publicKeyFileList, pubKey)
```

helpers/helpers.go
```diff
@@ -290,17 +290,16 @@ func GetAuth(path string) (*Config, error) {
...
 	}
 
-	if FileExists(".sda-cli-session") {
-		file, err := os.Open(".sda-cli-session")
-		if err != nil {
-			fmt.Println("could not read file:", file)
-		}
+	file, err := os.Open(".sda-cli-session")
+	if err != nil {
+		fmt.Println("could not read file:", file)
```
jbygdell, line 302:
```suggestion
		return "", err
```

helpers/helpers.go
```diff
@@ -314,13 +313,24 @@ func GetPublicKey() (string, error) {
 		return "", errors.New("public key not found in the configuration")
 	}
 
+	pubFile, err := CreatePubFile(config.PublicKey, "key-from-oidc.pub.pem")
+	if err != nil {
+		return "", fmt.Errorf("failed to create public key file: %w", err)
```
jbygdell, line 318:
```suggestion
		return "", err
```

helpers/testcreate.pub.pem
```diff
@@ -0,0 +1,3 @@
+-----BEGIN CRYPT4GH PUBLIC KEY-----
+MzM5ZWIyYTQ1OGZlYzVlMjNhYThiNTdjZmNiMzVmMTA=
+-----END CRYPT4GH PUBLIC KEY-----
```
jbygdell, line 3: Don't commit this file

helpers/helpers_test.go
```diff
@@ -370,10 +370,24 @@ public_key = 27be42445fd9e39c9be39e6b36a55e61e3801fc845f63781a813d3fe9977e17a
...
+	var expectedPubKey = `-----BEGIN CRYPT4GH PUBLIC KEY-----
+MzM5ZWIyYTQ1OGZlYzVlMjNhYThiNTdjZmNiMzVmMTA=
+-----END CRYPT4GH PUBLIC KEY-----
+`
+	_, err := CreatePubFile(pubKeyContent, "test_public_file.pub.pem")
+	assert.NoError(suite.T(), err)
+
+	pubFile, _ := os.ReadFile("test_public_file.pub.pem")
+	s := string(pubFile)
+	assert.Equal(suite.T(), expectedPubKey, s)
```
jbygdell, line 392: Here you need to remove the public key you wrote once the tests is completed.

login/login.go
```diff
@@ -34,7 +34,7 @@ var ArgHelp = `
     [login-target]
         The login target can be one of the following: 
 			https://login.bp.nbis.se/
-			https://login.test.fega.nbis.se/
+			https://login.fega.nbis.se/
```
jbygdell, line 37: remove all references to specific sites.

login/login.go
```diff
@@ -168,7 +168,7 @@ func NewLogin(args []string) error {
 	}
 	err = deviceLogin.Login()
 	if err != nil {
-		return fmt.Errorf("Login failed")
+		return fmt.Errorf("Login failed: %v", err)
```
jbygdell, line 171: Return the raw error and do the decoration in the calling function.

jbygdell (changes requested):

encrypt/encrypt.go
pahatz: There still should be some suggestion of how the target looks, or do you think that that information could be provided beforehand somewhere?

encrypt/encrypt.go
pahatz: No, I think we still need that, otherwise we will get in this block in the case of having set the config as an argument.

encrypt/encrypt.go
jbygdell: The target is an URL, we should not give explicit examples.

helpers/helpers_test.go
```diff
@@ -389,3 +389,17 @@ func (suite *HelperTests) TestInvalidCharacters() {
...
 	}
 }
+
+func (suite *HelperTests) TestCreatePubFile() {
+	var pubKeyContent = `339eb2a458fec5e23aa8b57cfcb35f10e7389025816e44d4234f814ed2aeed3f`
+	var expectedPubKey = `-----BEGIN CRYPT4GH PUBLIC KEY-----
+MzM5ZWIyYTQ1OGZlYzVlMjNhYThiNTdjZmNiMzVmMTA=
+-----END CRYPT4GH PUBLIC KEY-----
+`
+	_, err := CreatePubFile(pubKeyContent, "test_public_file.pub.pem")
```
jbygdell, line 399: Put this in the standard temp directory instead of the current working directory

encrypt/encrypt.go
jbygdell: Don't speculate.
If you are unsure, add a test case in the relevant test file.

encrypt/encrypt.go
```diff
@@ -26,7 +26,8 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help encrypt` command
 var Usage = `
-USAGE: %s encrypt -key <public-key-file> (-outdir <dir>) (-continue=true) [file(s)]
+USAGE: %s encrypt -key <public-key-file> (-target <target>) (-outdir <dir>) (-continue=true) [file(s)]
+
```
jbygdell, line 30:
```suggestion
```

jbygdell (changes requested):

encrypt/encrypt.go
pahatz: removed, not sure why github doesn't catch this

login/login.go
pahatz: I fixed this even though it still shows up, remote is up to date

helpers/helpers_test.go
pahatz: added

encrypt/encrypt.go
pahatz: My phrasing was misleading.
The clause is correct, the tests actually caught it.

jbygdell: There are no relevant test cases in the code right now.
And since `publicKeyFileList` is supposed to be `nil` here we don't need to test for it, if it hadn't been nil it would have been caught on line 82.

encrypt/encrypt.go
```diff
@@ -75,14 +77,38 @@ func Encrypt(args []string) error {
 		return err
 	}
 
-	// no key provided, check for one in the session file
-	if len(publicKeyFileList) == 0 {
+	var pubKey string
+
```
jbygdell, line 81:
```suggestion
```

This is not needed since the variable is only used inside the if block below

encrypt/encrypt.go
```diff
@@ -75,14 +77,38 @@ func Encrypt(args []string) error {
...
+			return err
+		}
+		log.Println("fetching public key")
+		// no key provided, no key in session file, target provided
+		publicKeyFileList = append(publicKeyFileList, pubKeyFile)
+	}
+	// no key provided, no key in session file, no target provided
+	if publicKeyFileList == nil && *target == "" {
+		// check for public key in .sda-cli-session file from login
+		pubKey, err = helpers.GetPublicKeyFromSession()
```
jbygdell, line 104:
```suggestion
		pubKey, err := helpers.GetPublicKeyFromSession()
```

Do full declaration here

jbygdell (approved):
Other than a dangling variable declaration it now looks ok.

login/login.go
pahatz: Is this coming from the two dot vs three dot diff that github does?
https://github.com/NBISweden/sda-cli/blob/264-fix-login-uris/login/login.go

jbygdell: No.
All URL needs to be removed.

login/login.go
```diff
@@ -32,10 +32,7 @@ login:
 // the module help
 var ArgHelp = `
     [login-target]
-        The login target can be one of the following: 
-			https://login.bp.nbis.se/
-			https://login.test.fega.nbis.se/
-			https://login.gdi.nbis.se/`
+        The login target is the base URL of the service.`
```
jbygdell, line 35: Your previous wording was better

jbygdell (approved):

aaperis (approved):
Looks good!
 I think that a short description on the order of reading-in the pubkey should be added in the `Usage` text of `encrypt` so that the user is aware of the expected behavior. For example:
```
$ ./sda-cli encrypt README.md 
Error: configuration file (.sda-cli-session) not found
```
does not seem super user-friendly without any context.

But this is not crucial and neither is my other comment.

encrypt/encrypt.go
```diff
@@ -75,14 +77,36 @@ func Encrypt(args []string) error {
...
 
-	// no key provided, check for one in the session file
-	if len(publicKeyFileList) == 0 {
+	if publicKeyFileList != nil && *target != "" {
+		return errors.New("only one of -key or -target can be used")
+	}
 
-		sesKey, err := helpers.GetPublicKey()
+	if *target != "" {
+		// fetch info endpoint values
```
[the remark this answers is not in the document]
pahatz: moved it

aaperis (approved):

jbygdell (approved):

PR #287, opened 2023-11-16
Disallow invalid characters in filepaths

### LINKED ISSUE #185: Disallow characters in filenames
For cross-os compatibility reasons we would like to not allow filenames with the following characters:

`\/:*?"<>|`

[see discussion here](https://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os).

MalinAhlberg:
This PR closes #185 .

Adds a check to see if there are invalid characters in any file path and reports an error if there is:
```sh
go run main.go upload -config ../sda-s3proxy/dev_utils/proxyS3 file\?path

Error: filepath file?path contains disallowed characters: ?
exit status 1
```

### DIFF
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/NBISweden/sda-cli/encrypt"
@@ -74,10 +73,9 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
 
 	// Loop through the list of file paths and check if their names are valid
 	for _, filename := range outFiles {
-		re := regexp.MustCompile(`[\\:\*\?"<>\|\x00-\x1F\x7F]`)
-		dissallowedChars := re.FindAllString(filename, -1)
-		if dissallowedChars != nil {
-			return fmt.Errorf("filepath %v contains disallowed characters: %+v", filename, strings.Join(dissallowedChars, ", "))
+		err := helpers.CheckValidChars(filename)
+		if err != nil {
+			return err
 		}
 	}
 
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -261,14 +261,19 @@ func Upload(args []string) error {
 		return fmt.Errorf("failed parsing arguments, reason: %v", err)
 	}
 
-	// Check that specified target directory is valid, i.e. not a filepath or a flag
-	info, err := os.Stat(*targetDir)
-
 	// Dereference the pointer to a string
 	var targetDirString string
 	if targetDir != nil {
 		targetDirString = *targetDir
 	}
+
+	err = helpers.CheckValidChars(filepath.ToSlash(targetDirString))
+	if err != nil {
+		return err
+	}
+
+	// Check that specified target directory is valid, i.e. not a filepath or a flag
+	info, err := os.Stat(*targetDir)
 
 	if (!os.IsNotExist(err) && !info.IsDir()) || (targetDirString != "" && targetDirString[0:1] == "-") {
 		return errors.New(*targetDir + " is not a valid target directory")
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -235,7 +235,7 @@ func createFilePaths(dirPath string) ([]string, []string, error) {
 			// Remove possible trailing "/" so that "path" and "path/" behave the same
 			dirPath = strings.TrimSuffix(dirPath, string(os.PathSeparator))
 			pathToTrim := strings.TrimSuffix(dirPath, filepath.Base(dirPath))
-			outPath := formatUploadFilePath(strings.TrimPrefix(path, pathToTrim))
+			outPath := filepath.ToSlash(strings.TrimPrefix(path, pathToTrim))
 			outFiles = append(outFiles, outPath)
 		}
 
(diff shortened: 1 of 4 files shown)


upload/upload_test.go
```diff
@@ -135,21 +135,6 @@ func (suite *TestSuite) TestcreateFilePaths() {
...
-	uploadPath := formatUploadFilePath(testPath)
-
-	assert.Equal(suite.T(), unixPath, uploadPath)
-
-	weirdPath := "a/b:/c;.c4gh"
-	goodPath := "a/b_/c_.c4gh"
-	formattedPath := formatUploadFilePath(weirdPath)
-
-	assert.Equal(suite.T(), goodPath, formattedPath)
-}
```
jbygdell, line 151: This test should be replaced by one where ensure files with invalid characters result in an error

MalinAhlberg: I moved out the check for invalid chars to [helpers](https://github.com/NBISweden/sda-cli/pull/287/files#diff-ba41f3a7938a59c0335de7decc4304f93218d516ce4b6340665a6e7b001e6230R402) so that it can be tested on its own. Running [`upload`](https://github.com/NBISweden/sda-cli/pull/287/files#diff-3ba32a4f84fefb210b231b3311ecc0af4f2805b0f7278ee53c1f4044ad2f603cR490) is only tested for linux, since windows won't allow having these characters in a filepath, and the test will hence fail on the fail creation.

upload/upload_test.go
```diff
@@ -438,3 +423,74 @@ func (suite *TestSuite) TestRecursiveToDifferentTarget() {
 
 	log.SetOutput(os.Stdout)
 }
+
+func (suite *TestSuite) TestUploadInvalidCharacters() {
+	// Filenames with :\?* can not be created on windows
+	if runtime.GOOS == "windows" {
+		return
```
jbygdell, line 430:
```suggestion
		suite.T().Skip()
```

You should be able to use something like the `skip` in order to show that the test where skipped rather than give the impression that it passed.

MalinAhlberg: Good point! Done in 260f1d8:

2023-11-28-160019_839x36_scrot

aaperis:
It looks good! ...but if you want to do it aaall the way, you need to treat the `-targetDir` option as well:
```
/sda-cli upload -config testing/s3cmd.conf -force-unencrypted LICENSE -targetDir a?d/d
The provided token expires in less than 24 hours
Consider renewing the token.
Input file LICENSE is not encrypted
force-unencrypted flag provided, continuing...
Uploading LICENSE with config testing/s3cmd.conf
File LICENSE: uploading 33.7 KiB / 33.7 KiB [========================================================================================================================]  done
Error: NotAcceptable: Not Acceptable
        status code: 406, request id: , host id: 
```

The above was captured by the `s3inbox` but not by the `sda-cli`. This case applies to windows users of the `sda-cli` as well, since `targetDir` is just a string.

jbygdell:
> The above was captured by the `s3inbox` but not by the `sda-cli`. This case applies to windows users of the `sda-cli` as well, since `targetDir` is just a string.

This here is a prime reason to do some serious refactoring of this code base.

There have been to much ad-hoc additions with no real thought behind the UX other than to get a specific functionality in.

Some issues found so far:
* running `if exists X` followed directly by a `if not exists X` 
* global variables being reset at the start of a function so that the test should pass.
* Check for flag options not being contained to sub functions or isolated blocks
* flag options not checked or enforced
* contra indicative flags options allowed in the same context.

I have started to create issues for these findings.

aaperis:
> > The above was captured by the `s3inbox` but not by the `sda-cli`. This case applies to windows users of the `sda-cli` as well, since `targetDir` is just a string.
> 
> This here is a prime reason to do some serious refactoring of this code base.
> 
> There have been to much ad-hoc additions with no real thought behind the UX other than to get a specific functionality in.
> 
> Some issues found so far:
> 
>     * running `if exists X` followed directly by a `if not exists X`
> 
>     * global variables being reset at the start of a function so that the test should pass.
> 
>     * Check for flag options not being contained to sub functions or isolated blocks
> 
>     * flag options not checked or enforced
> 
>     * contra indicative flags options allowed in the same context.
> 
> 
> I have started to create issues for these findings.

Finding ways to improve the code is always great and refactoring has never been done in this case as far as I remember. This is something definitely needed.

However, I don't think that this lengthy message here is relevant to the actual discussion of this PR.

MalinAhlberg:
Thanks for the reviews @jbygdell and @aaperis, good that you checked those things (and yes, let's go aaaall the way :muscle: (at least all the way to here :sweat_smile: ) ).

I have added commit d75b50c for the warning/panic messages.
c7b4986 and e6b6639 add a check of the targetDir, together with some tests.

Great that you created issues for the refactoring @jbygdell :star: !

upload/upload.go
```diff
@@ -71,6 +71,14 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
 		return errors.New("no files to upload")
 	}
 
+	// Loop through the list of file paths and check if their names are valid
+	for _, filename := range outFiles {
+		err := helpers.CheckValidChars(filename)
```
jbygdell, line 76: we should only have to do this check once regardless if `targetDir` is specified or not, but that means refactoring the entire upload function.
#save-for-later

jbygdell (approved):

PR #299, opened 2023-12-04
Bugfix/fail fast if token is expired

### LINKED ISSUE #295: No error if the token in the config file has expired
compare the exp date in the token to the current date a bit better.

a/c
fail hard if the token has expired

aaperis:
This PR fixes #295 but also refactors how token expiration checking is performed and reported to the user.

Main changes:
- The `sda-cli` will exit with error upon detecting an expired  access token. 
- The `sda-cli` will now always print the remaining token life in case its `exp` date is still valid 
- In case the token is valid but the `exp` date is less than a day, it will also print a reminder to renew the token
- Test tokens are now generated  on the fly and hard coded ones are removed the from the test suite


helpers/helpers_test.go
```diff
@@ -1,13 +1,17 @@
...
+	"crypto/rsa"
 	"fmt"
 	"log"
 	"os"
 	"runtime"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/golang-jwt/jwt"
```
jbygdell, line 14:
```suggestion
	""github.com/lestrrat-go/jwx/v2/jwt""
```

Since you are doing a refactoring. Consider using the same JWT library used elsewhere in the code suite, it will make the merging easier down the road.

helpers/helpers.go
```diff
@@ -343,34 +344,54 @@ func GetPublicKey() (string, error) {
...
+	switch iat := claims["exp"].(type) {
+	case float64:
+		expiration = time.Unix(int64(iat), 0)
+	case json.Number:
+		tmp, _ := iat.Int64()
+		expiration = time.Unix(tmp, 0)
+	case string:
+		i, err := strconv.ParseInt(iat, 10, 64)
+		if err != nil {
+			return fmt.Errorf("could not parse token, reason: %s", err)
```
jbygdell, line 377: This should not really be needed.
The `"github.com/lestrrat-go/jwx/v2/jwt"` library will do this validation for you.

jbygdell (dismissed):
Consider using `github.com/lestrrat-go/jwx/v2/jwt` for all JWT handling since that is the one used elsewhere. 
But that could easily be another PR, approving as is.

helpers/helpers.go
aaperis: Yeah, this looked a bit weird not to be handled by the library. I'll take a look at `jwx` library.

upload/upload.go
```diff
@@ -138,7 +138,7 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
 		}
 		fileExists, err := helpers.ListFiles(*config, listPrefix)
 		if err != nil {
-			log.Error("Couldn't get the file list ", err)
+			return fmt.Errorf("listing uploaded files: %s", err.Error())
```
nanjiangshu, line 141: Just curious, in which condition will this error be triggered?

nanjiangshu (dismissed):
Looks great! All points mentioned in the PR description seems covered.

upload/upload.go
aaperis: Before the PR `sda-cli` would proceed with an upload regardless of an expired token. However, since the token was expired, the call to `helpers.ListFiles` would return an error because the user could not be authenticated. Then, the error would be `logged` instead of propagated, causing the `sda-cli` to continue and  `panic` somewhere down the road before the tool would finally exit.

aaperis:
> Consider using `github.com/lestrrat-go/jwx/v2/jwt` for all JWT handling since that is the one used elsewhere. But that could easily be another PR, approving as is.

I agree that we should change the library for `jwt` handling to the above and improve parts of the logic using its capabilities. I am deferring this for now by opening a refactoring issue for this point so that we can merge this fast.

upload/upload.go
```diff
@@ -138,7 +138,7 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
 		}
 		fileExists, err := helpers.ListFiles(*config, listPrefix)
 		if err != nil {
-			log.Error("Couldn't get the file list ", err)
```
pahatz, line 141: It would also be a good idea to change the order of doing things here below.
Now we call `ListFiles` first and then check if the `forceOverwrite` flag is set.
We could move the check before calling `ListFiles`. This way if the flag is set, we avoid making that call.
If you think this should be another issue just, disregard the comment.

pahatz (approved):
Looks great!

upload/upload.go
```diff
@@ -129,26 +129,23 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
...
+			}
+			if len(fileExists.Contents) > 0 && aws.StringValue(fileExists.Contents[0].Key) == filepath.Clean(config.AccessKey+"/"+targetDir+"/"+outFiles[k]) {
 				fmt.Printf("File %s is already uploaded!\n", filepath.Base(filename))
-				if !*forceOverwrite {
-					fmt.Println("Quitting...")
 
-					return errors.New("file already uploaded")
-				}
-				fmt.Println("force-overwrite flag provided, continuing...")
+				return errors.New("file already uploaded")
```
jbygdell, line 148: How are these changes related to the token expiring?

jbygdell (dismissed):
Drop the last commit

upload/upload.go
aaperis: Admittedly, they're not.

jbygdell: Then it could be included in a scoped refactoring PR (targeting only the upload functionality).
Since it is not remotely related to this PR it should not be here.

jbygdell (approved):

PR #309, opened 2023-12-14
two decoder links added

betuleren:
This PR closes [#301](https://app.zenhub.com/workspaces/cortex-6374c04dd78d631b359d99e5/issues/gh/nbisweden/aida-data-hub/301).

**Description**
Two decoder links have been added to README.md.

**How to test**

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ To get help on the usage of the tool, please use the following command
 
 The main functionalities implemented in this tool are explained in the following sections.
 
+If any command seems a bit puzzling, you can find guidance on how to interpret it right [here](https://ftpdocs.broadcom.com/cadocs/0/CA%20ARCserve%20%20Backup%2015-ENU/Bookshelf_Files/HTML/CMD_Ref/command_line_syntax_characters.htm) and [here](https://medium.com/@jaewei.j.w/how-to-read-man-page-synopsis-3408e7fd0e42).
+
 ## Encrypt
 
 The files stored in the SDA/BP archive are encrypted using the [crypt4gh standard](https://www.ga4gh.org/news/crypt4gh-a-secure-method-for-sharing-human-genetic-data/). The following sections explain how to encrypt and upload files to the archive.


pontus (approved):
Looks good to me, thanks!

pahatz:
I am not sure if we should have external links inside the sda-cli tool itself for a number of reasons.
They are external, we don't maintain or control them, they might as well get compromised so it's a potential risk.
 sda-cli tool could be sandboxed from the internet.
I could see it as part of a guide we hand over to users who aren't familiar with command line tools.
`decoder links` is also a bit misleading, I would refrain from using the term to avoid confusion with the algorithmic decoding.

pontus:
> I am not sure if we should have external links inside the sda-cli tool itself for a number of reasons. They are external, we don't maintain or control them, they might as well get compromised so it's a potential risk. sda-cli tool could be sandboxed from the internet. I could see it as part of a guide we hand over to users who aren't familiar with command line tools. `decoder links` is also a bit misleading, I would refrain from using the term to avoid confusion with the algorithmic decoding.

I think this PR was mostly created because of a request from me, so I feel I should comment on this. I agree with the facts on linking externally (can break, become compromised, et.c.) but consider the risks with that fairly low (if/when that happens, we can act on that).

Given that we're pointing end users to this repository, I definitely want some sort of "explainer" - are target group isn't only comprised of people used to Unixes and I've had questions because it wasn't understandable. 

Given that we currently don't want to commit to writing a full-sized sda-cli guide for the larger target groups (or a guide of our own on how to interpret man-page syntax), something like this feels like a large improvement for our users for little effort on our part. Obviously fine to disagree, but even if we go for any such more work-intensive option, I'd still wish for something like this as a stopgap solution unless someone has strong objections.

I can understand if there can be issues with the term "decoder links" but don't see it as a large problem as it's only the branch/PR name, not actual content.

aaperis (approved):

PR #319, opened 2024-01-05
Refactor decrypt function

jbygdell:
**Description**
* always prompt for passphrase unless ENV `C4GH_PASSWORD` is set
* add `--force-overwrite` option
* Expand the test suite to cover logical user actions
* code simplification

**How to test**
sda-cli decrypt -key KEY FILE
sda-cli decrypt -key KEY --force-overwrite FILE

### DIFF
--- a/decrypt/decrypt.go
+++ b/decrypt/decrypt.go
@@ -84,9 +84,16 @@ func Decrypt(args []string) error {
 	}
 
 	// Check that all the encrypted files exist, and all the unencrypted don't
-	err = checkFiles(files)
-	if err != nil {
-		return err
+	for _, file := range files {
+		// check that the input file exists and is readable
+		if !helpers.FileIsReadable(file.Encrypted) {
+			return fmt.Errorf("cannot read input file %s", file.Encrypted)
+		}
+
+		// check that the output file doesn't exist
+		if helpers.FileExists(file.Unencrypted) {
+			return fmt.Errorf("outfile %s already exists", file.Unencrypted)
+		}
 	}
 
 	// decrypt the input files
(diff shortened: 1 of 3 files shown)


decrypt/decrypt_test.go
```diff
@@ -75,91 +60,99 @@ func (suite *DecryptTests) TestreadPrivateKeyFile() {
...
 	// Test decrypting a non-existent file
-	err = decrypt(filepath.Join(suite.tempDir, "non-existent"), "output_file", *privateKey)
-	assert.EqualError(suite.T(), err, fmt.Sprintf("infile %s does not exist or could not be read", filepath.Join(suite.tempDir, "non-existent")))
-
-	// Test decrypting where the output file exists
-	err = decrypt(encryptedFile, suite.testFile.Name(), *privateKey)
-	assert.EqualError(suite.T(), err, fmt.Sprintf("outfile %s already exists", suite.testFile.Name()))
+	msg := "no such file or directory"
+	if runtime.GOOS == "windows" {
+		msg = "The system cannot find the file specified."
```
pahatz, line 103: Where does this specific message come from for the case of windows OS?

pahatz (approved):
Looks good overall.

decrypt/decrypt.go
```diff
@@ -69,37 +65,38 @@ func Decrypt(args []string) error {
...
-		password, err := getPassword("C4GH_PASSWORD")
-		if err != nil {
-			return err
-		}
-
-		// Loading private key file
-		privateKey, err = readPrivateKeyFile(*privateKeyFile, password)
+	password, available := os.LookupEnv("C4GH_PASSWORD")
+	if !available {
+		password, err = helpers.PromptPassword("Enter password to unlock private key")
```
aaperis, line 70: This makes the tool less user friendly imo.

jbygdell: As long as the user password protects the private key there is no difference in the behavior of the tool.

pontus (approved):

Issue #314, opened 2023-12-18
Add tests for cases when `sda-cli` calls the /info endpoint of `sda-auth` 


aaperis:
**Please describe the feature**
As a developer I want test automation so that I am sure the integration between `sda-cli` and `sda-auth` works as expected with every added feature.

**Acceptance criteria**
- [ ] Have a mock `info` endpoint  
- [ ] Test the order of reading in c4gh pubkey for the encrypt module

**Additional context**

PR #344, opened 2024-02-23
Fix/device login

### LINKED ISSUE #329: Client login with sda-cli
As an sda-dev
I want the sda-cli to be able to login with LS-AAI
In order to make it easier for users to get a token

The sda-cli cannot connect to an external login URL

aaperis:
This PR closes #329.

**Description**

The reason why `sda-cli login` didnot seem to work is because it tried to connect to a `/token` that was behind a basic authentication method.

As a public app, `sda-cli` cannot leverage traditional methods to authenticate to the `/token` endpoint of LS-AAI.  

- To mitigate this issue, this PR adds to the device flow a PKCE flow so that the `sda-cli` can connect securely to  LS-AAI token endpoints that do not require authentication through sending secrets.

- Additionally, a `mock-oidc` server configured for device-flow and `sda-auth` configured only for serving the `/info` endpoint are added to the docker-compose.yml of the local dev environment for local testing. These additions can be the starting point for work in 
#314 
and 
#252. 

- along the way go was updated to 1.21 (crypt4gh v1.8.11 required this)

**How to test**
Configure an OIDC service with device flow enabled,  token_authentication_method=none and PKCE enabled with method=S256.

OR locally:
under `/testing` folder run:
```
export TAG=v0.2.103 && docker compose --profile login up auth oidc --build --force-recreate
```
and from another terminal:
```
go build .
./sda-cli login http://localhost:8080
```
and after the login succeeds `ls .sda-cli-session` .


pahatz (approved):
Looks good!

pontus (dismissed):

testing/docker-compose.yml
```diff
@@ -101,6 +101,59 @@ services:
...
+      retries: 4
+  keygen:
+    profiles: ["login"]
+    image: golang:alpine3.16
+    container_name: keygen
+    command:
+      - "/bin/sh"
+      - "-c"
+      - if [ ! -f "/out/c4gh.sec.pem" ]; then wget -qO- "https://github.com/neicnordic/crypt4gh/releases/latest/download/crypt4gh_linux_x86_64.tar.gz" | tar zxf -;
+        ./crypt4gh generate -n c4gh -p privatekeypass && mv *.pem /out/; fi
```
jbygdell, line 133:
```suggestion
      - if [ ! -f "/out/c4gh.sec.pem" ]; then wget -qO- "https://github.com/neicnordic/crypt4gh/releases/latest/download/crypt4gh_linux_x86_64.tar.gz" | tar zxf -;
        ./crypt4gh generate -n /out/c4gh -p privatekeypass; fi
```

testing/docker-compose.yml
```diff
@@ -101,6 +101,59 @@ services:
...
+    volumes:
+      - /tmp:/out
```
jbygdell, line 135: Not really a fan of this, yes it is written o `/tmp` but the files are still owned by root since the container doesn't have a `user:` stansa.

testing/oidc/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY package.json ./
+
+RUN  rm -rf node_modules
+
+RUN npm install -g npm@latest && \
+    npm i camelcase oidc-provider
+
+COPY . .
+
+CMD [ "node", "server.js" ]
```
jbygdell, line 16: Why not use the `ls-aai-mock` from GDI?

jbygdell (approved):
Looks reasonable, some comments and questions though.

testing/docker-compose.yml
aaperis: Not fan of this either. Changed it to a named volume :-)

testing/oidc/Dockerfile
aaperis: Good point! The GDI repo does not provide an image for ls-aai-mock, so I thought the current solution is better than pulling that repo and building images as it can be used for integration tests in the future. But, I can see if I can add some README instructions on how to plug in the ls-aai-mock as well for further testing.

jbygdell: The only thing you need is in the ls-aai-mock docker compose file.

login/login.go
```diff
@@ -161,7 +166,7 @@ func (login *DeviceLogin) UpdateConfigFile() error {
 func NewLogin(args []string) error {
 	deviceLogin, err := NewDeviceLogin(args)
 	if err != nil {
-		return fmt.Errorf("failed to contact authentication service")
```
MalinAhlberg, line 164: Just checking: could the now left out information about where the error was raised not be good to have for debugging? Eg `failed to contact authentication service`, `failed parsing arguments` etc.

MalinAhlberg (approved):
Looks good!
I left one minor question/comment. Also, I needed to run `go mod tidy` before being able to run `go build`.

kostas-kou (commented):
I agree with Malin's comment. Easier to debug when one is logging their own error messages as well

login/login.go
aaperis: That is a good point, however I think the returned error is xplanatory enough so that this text is not needed. If I am not mistaken this is something like:
```
./sda-cli login http://lalala
Error: Get "http://lalala/info": dial tcp: lookup lalala: no such host
```

pontus: I also agree with Malin. I'd consider it best practice to include source error messages unless there is a security concern (e.g. risk of e.g. information leak).

Since this is on the client, no such risk.

(I also note that e.g. the "no such host" error shown above can be vital information point in debugging, e.g. in an enterprise system with broken DNS, or of course "no route to host" or "connection refused" would also be important parts in figuring out what goes wrong. I've actually heard that there are people that do not like starting their debugging with `strace` to figure out what system calls are made and what they return.)

login/login.go
```diff
@@ -179,14 +184,14 @@ func NewDeviceLogin(args []string) (DeviceLogin, error) {
 	var url string
 	err := Args.Parse(args[1:])
 	if err != nil {
-		return DeviceLogin{}, errors.New("failed parsing arguments")
+		return DeviceLogin{}, err
```
pontus, line 187: As above, passing errors onwards is nice, I generally prefer making a new error (including the old one) to clarify where the current comes from, e.g. `fmt.Errorf("Arguments parsing failed: %v, err)`.

login/login.go
```diff
@@ -218,7 +223,7 @@ func (login *DeviceLogin) Login() error {
 	var err error
 	login.wellKnown, err = login.getWellKnown()
 	if err != nil {
-		return fmt.Errorf("failed to fetch .well-known configuration: %v", err)
+		return err
```
pontus, line 226: I have preference for keeping.

```suggestion
		return fmt.Errorf("failed to fetch .well-known configuration: %v", err)
```

kostas-kou (approved):

aaperis:
> It works great when I tested locally with the instructions provided! I have one minor comment though When the URL has a trailing character `/`, which may happen when a user copies the URL from the address bar, the error message reported is not meaningful. For example, running the command
> 
> ```
> ./sda-cli login https://login.test.bp.nbis.se/
> ```
> 
> will return `Error: invalid character 'N' looking for beginning of value`
> 
> One can either sanitize the trailing `/` from the URL or provide a more understandable error message.

aaperis:
> It works great when I tested locally with the instructions provided! I have one minor comment though When the URL has a trailing character `/`, which may happen when a user copies the URL from the address bar, the error message reported is not meaningful. For example, running the command
> 
> ```
> ./sda-cli login https://login.test.bp.nbis.se/
> ```
> 
> will return `Error: invalid character 'N' looking for beginning of value`
> 
> One can either sanitize the trailing `/` from the URL or provide a more understandable error message.

fixed

aaperis:
Apparently `merge_queue` expects two tests for golang 1.20 to finish but these are now run as 1.21 , It seems that I don't have enough access to modify the relevant settings at least atm.

login/login.go
aaperis: I changed back the extra error messages and added the actual err with them.

README.md
```diff
@@ -235,7 +235,7 @@ You can get the current version of the sda-cli by running:
 This section contains the information required to install, modify and run the `sda-cli` tool.
 
 ## Requirements
-The `sda-cli` is written in golang. In order to be able to modify, build and run the tool, golang (>= 1.20) needs to be installed. The instructions for installing `go` can be found [here](https://go.dev/doc/install).
+The `sda-cli` is written in golang. In order to be able to modify, build and run the tool, golang (>= 1.21) needs to be installed. The instructions for installing `go` can be found [here](https://go.dev/doc/install).
```
pontus, line 238:
```suggestion
The `sda-cli` is written in golang. In order to be able to modify and build the tool, golang (>= 1.21) needs to be installed. The instructions for installing `go` can be found [here](https://go.dev/doc/install).

Binaries built do not require the golang environment and can be moved to other machines with the same OS and processor architecture and run there.
```

Also clarify what can be interpreted as all users needing to have golang installed?

pontus (approved):

README.md
aaperis: This part is under the `Developer's section`, so this applies to developers not all users. Not sure what needs to be clarified further :-)

pahatz (approved):

PR #379, opened 2024-05-15
Clarify the instructions for uploading

### LINKED ISSUE #368: Misleading command instruction for `sda-cli upload`
**Describe the bug**
The command instruction of  
```bash
sda-cli help upload
```
is misleading 
- The general description 
    ```plaintext
    Uploads files to the Sensitive Data Archive (SDA).  All files
    to upload are required to be encrypted and have the .c4gh file
    extension.
    ```  
     is not very accurate.
     1. . It's not true that the file to upload must have `.c4gh` extension, one can rename a c4gh encrypted file to e.g. `file.data` and it can still be uploaded.  
     2. Non encrypted file can also be uploaded when the flag `--encrypt-with-key` is set. Although this is mentioned in the instruction for that flag, it is a bit misleading here.

 
- The instruction for 
     ```plaintext  
     -encrypt-with-key string
    	Public key file to use for encryption of files before upload.
    	The key file may optionally contain several concatenated
    	public keys.  The argument list may include only unencrypted
    	data if this flag is set.
    ```
    is not very accurate. When this flag is set, only unencrypted files should be provided, `may` should be `must`.

**Expected behavior**
-  Suggested changes for the general description:

```plaintext
All files to upload are required to be crypt4gh encrypted when the flag --encrypt-with-key is not set, 
or non encrypted when the flag --encrypt-with-key is set
```

- Suggested changes for the flag `--encrypt-wit-key`

     ```plaintext
      -encrypt-with-key string
    	Public key file to use for encryption of files before upload.
    	The key file may optionally contain several concatenated
    	public keys.  Only unencrypted data should be provided when this flag is set.
    ```

**Additional context**

**Estimation of size**: small

nanjiangshu:
This PR closes #368 

**Description**
The instructions for `sda-cli upload` should be clear.

### DIFF
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -28,9 +28,9 @@ var Usage = `
 USAGE: %s upload -config <s3config-file> (--encrypt-with-key <public-key-file>) (--force-overwrite) (--force-unencrypted) (-r) [file(s) | folder(s)] (-targetDir <upload-directory>)
 
 upload:
-    Uploads files to the Sensitive Data Archive (SDA).  All files
-    to upload are required to be encrypted and have the .c4gh file
-    extension.
+    Uploads files to the Sensitive Data Archive (SDA).
+    All files to upload are required to be crypt4gh encrypted when the flag -encrypt-with-key is not set.
+    On the other hand, non-encrypted files should be provided when the flag -encrypt-with-key is set.
 `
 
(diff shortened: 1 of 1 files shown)


MalinAhlberg (approved):
Nice!

aaperis (approved):

PR #380, opened 2024-05-17
Fix goreleaser so we get automatic builds back

### LINKED ISSUE #365: Fix releaser
This looks like a bug coming from this: https://github.com/NBISweden/sda-cli/commit/e111be729947e0a6d84acb6d15742f8bb94f5a5a where `GORELEASER_CURRENT_TAG` does not seem to be valid. 
logs: https://github.com/NBISweden/sda-cli/actions/runs/8786298524/job/24108852058

pontus:
This should make the goreleaser workflow function again.

Goreleaser seems to have both `.Version` supposedly meaning the version derived, but without initial `v` in our case, and `.Tag` meaning the git tag (see https://goreleaser.com/customization/templates/) for more information). For consistency, I think `.Tag` is what we want, but could to with `.Version` as well if preferred.

Closes #365 .

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -15,7 +15,7 @@ builds:
       - goarch: 386
       - goarch: arm
     ldflags:
-      - -s -w -X main.Version={{ GORELEASER_CURRENT_TAG }}
+      - -s -w -X main.Version={{ .Tag }}
 archives:
   - name_template: >-
       {{ .ProjectName }}_


aaperis (approved):
I am ok with either, I am just thinking how we would like `./sda-cli-version` to work. It seems we hardcode the version at the moment but we need to be consistent here I think.

kostas-kou (approved):

PR #392, opened 2024-05-30
Feature/sda download

### LINKED ISSUE #372: Use sda-download's api to get files
As an sda-admin
I want to have a tool that allows users to download files that they have access to
so that I don't have to manually publicize them.

Description:
`sda-cli` needs to know which download service api to talk to. The service api URI can be passed as an argument for this issue.

A/C:
- Given a token with the correct visa for a dataset, download a file from `/files` endpoint (or the relevant file serving endpoint for each specific case) of sda-download.
- ~The downloaded file is encrypted with the user's key.~

Demo;
- download file

kostas-kou:
This PR closes #372.

**Description**
A new feature in the sda-cli is implemented to download files from the download service endpoint by giving the dataset stable_id and the inbox path of the file without the uploader's user id

**How to test**
By testing that just run the setup script from the root folder: `bash .github/integration/setup/setup.sh`
and when it is done one can download the file by running: `./sda-cli sda-download -config testing/s3cmd-download.conf -dataset https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 -outdir test-download main/subfolder/dummy_data.c4gh`
Then check if the file `test-download/main/subfolder/dummy_data` exists and if the result by running `head test-download/main/subfolder/dummy_data` is "THIS FILE IS JUST DUMMY DATA"


sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+		if strings.Contains(file.FilePath, filename) {
+			fileID = file.FileID
+
+			break
+		}
+	}
+
+	if fileID == "" {
+		return "", fmt.Errorf("failed to find file ID for %s", filename)
+	}
```
jbygdell, line 222: You don't need to iterate through the list to find the entry you want.
something like this should solve it:
```go
idx := slices.IndexFunc(myconfig, func(f files) bool { return f.FileName == filename })
if idx == -1 {
    return "", fmt.Errorf("failed to find file ID for %s", filename)
}

then use `files[idx]` in the return statement at the end.
```

This should be expanded so that the user can supply either a file ID or a file path.

sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Wait for the progress bar to finish
+	p.Wait()
+
+	return nil
+}
+
+// downloadURL gets the datset files, parses the JSON response to get the file ID
+// and returns the download URL for the file
+func downloadURL(baseURL, token, dataset, filename string) (string, error) {
```
jbygdell, line 187: This should have a better name, something like `getFileIDURL`

sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/vbauerster/mpb/v8"
+	"github.com/vbauerster/mpb/v8/decor"
+)
+
+// Help text and command line flags.
+
+// Usage text that will be displayed as command line help text when using the
+// `help download` command
+var Usage = `
+USAGE: %s sda-download -config <s3config-file> -dataset <datasetID> -url <uri> (-outdir <dir>) [filename(s)]
```
jbygdell, line 24: We should make the cli read the token from ENV since the config file is mainly for uploading data and that can hav ea resigned token.

main.go
```diff
@@ -70,6 +72,8 @@ func main() {
 		err = list.List(args)
 	case "login":
 		err = login.NewLogin(args)
+	case "sda-download":
+		err = sdaDownload.SdaDownload(args)
```
jbygdell, line 76: This shouldn't really  be a new command, we should just have `download` since the current download from exported s3 buckets will be deprecated.

sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+		All flagless arguments will be used as sda-download uri.
+	[filename(s)]
+		The name of the file to download.`
+
+// Args is a flagset that needs to be exported so that it can be written to the
+// main program help
+var Args = flag.NewFlagSet("sda-download", flag.ExitOnError)
+
+var configPath = Args.String("config", "",
+	"S3 config file to use for downloading.")
```
jbygdell, line 47:
```suggestion
var configPath = Args.String("config", "", "S3 config file to use for downloading.")
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+
+var datasetID = Args.String("dataset", "",
+	"Dataset ID for the file to download")
```
jbygdell, line 50:
```suggestion
var datasetID = Args.String("dataset", "", "Dataset ID for the file to download")
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+
+var url = Args.String("url", "",
+	"The name of the file to download")
```
jbygdell, line 53:
```suggestion
var url = Args.String("url", "", "The name of the file to download")
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+
+var outDir = Args.String("outdir", "",
+	"Directory for downloaded files.")
```
jbygdell, line 56:
```suggestion
var outDir = Args.String("outdir", "", "Directory for downloaded files.")
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,259 @@
...
+	for _, filePath := range files {
+		downloadurl, err := downloadURL(*url, config.AccessToken, *datasetID, filePath)
+		if err != nil {
+			return err
+		}
+
+		filePathSplit := strings.Split(filePath, "/")
+		if strings.Contains(filePath, "elixir-europe.org") {
+			filePath = strings.Join(filePathSplit[1:], "/")
+		}
```
jbygdell, line 120: This can't be hard coded

jbygdell (changes requested):
* The app should only have one download entry
* The static tests mixes error catching and assertions.
* There is a lot of excess whitespace in the form on empty lines directly after function and variable declarations.

sda_download/sda_download.go
kostas-kou: That is cool, I had no clue about the IndexFunc.
Done

sda_download/sda_download.go
kostas-kou: Done

sda_download/sda_download.go
kostas-kou: done

sda_download/sda_download.go
kostas-kou: Done

sda_download/sda_download.go
kostas-kou: Done

sda_download/sda_download.go
kostas-kou: Done

sda_download/sda_download.go
kostas-kou: Done

sda_download/sda_download.go
kostas-kou: Yes we can do that but on the other hand I think that is easier for the common user to to just download a file and then not bother about anything else. We can discuss it tomorrow with the rest of the team.

main.go
kostas-kou: Correct but as far as I remember we will keep both for some time and when the other one will be deprecated then for sure we can change the name

sda_download/sda_download.go
kostas-kou: In that part I am just checking if the userid is in the filepath in order not to remove and not making it part of the download path. Unless you mean that even that information is too much to exist there

kostas-kou:
The tests are broken after the push of the last commit which was resolving some of the Jocke's comments. Again locally everything runs but not in actions.  I am trying to fix that again

sda_download/sda_download.go
jbygdell: I mean the `elixir-europe.org` part. Because that will change depending on which instance you authenticate to.

jbygdell: That will fail in FEGA where the user ID is of the format USER.NAME@HOST

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// getFileIDURL gets the datset files, parses the JSON response to get the file ID
+// and returns the download URL for the file
+func getFileIDURL(baseURL, token, dataset, filename string) (string, error) {
+	// Sanitize the base_url
+	baseURL = strings.TrimSuffix(baseURL, "/")
+	if !strings.HasPrefix(baseURL, "http") {
+		return "", fmt.Errorf("invalid URL, missing protocol (http/https)")
+	}
```
jbygdell, line 198:
```suggestion
	u, err := url.ParseRequestURI(baseURL)
	if err != nil || u.Scheme == "" {
		return "", err
	}
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
...
+
+// Args is a flagset that needs to be exported so that it can be written to the
+// main program help
+var Args = flag.NewFlagSet("sda-download", flag.ExitOnError)
+
+var configPath = Args.String("config", "", "S3 config file to use for downloading.")
+
+var datasetID = Args.String("dataset", "", "Dataset ID for the file to download")
+
+var url = Args.String("url", "", "The name of the file to download")
```
jbygdell, line 52:
```suggestion
var URL = Args.String("url", "", "The name of the file to download")
```

Don't user variable names that conflict with standard libraries

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
...
+// download's service APIs
+func SdaDownload(args []string) error {
+	var files []string
+	// Call ParseArgs to take care of all the flag parsing
+	err := helpers.ParseArgs(args, Args)
+	if err != nil {
+		return fmt.Errorf("failed parsing arguments, reason: %v", err)
+	}
+
+	if *datasetID == "" || *url == "" || *configPath == "" {
```
jbygdell, line 85:
```suggestion
	if *datasetID == "" || *URL == "" || *configPath == "" {
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
...
+
+	// Check if the token has expired
+	err = helpers.CheckTokenExpiration(config.AccessToken)
+	if err != nil {
+		return err
+	}
+
+	// Loop through the files and download them
+	for _, filePath := range files {
+		fileIDURL, err := getFileIDURL(*url, config.AccessToken, *datasetID, filePath)
```
jbygdell, line 110:
```suggestion
		fileIDURL, err := getFileIDURL(*URL, config.AccessToken, *datasetID, filePath)
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
+package sdadownload
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"net/http"
```
jbygdell, line 9:
```suggestion
	"net/http"
	"net/mail"
	"net/url"
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
```
jbygdell, line 12:
```suggestion
```

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
...
+// and returns the download URL for the file
+func getFileIDURL(baseURL, token, dataset, filename string) (string, error) {
+	// Sanitize the base_url
+	baseURL = strings.TrimSuffix(baseURL, "/")
+	if !strings.HasPrefix(baseURL, "http") {
+		return "", fmt.Errorf("invalid URL, missing protocol (http/https)")
+	}
+
+	// Make the url for listing files
+	filesURL := baseURL + "/metadata/datasets/" + dataset + "/files"
```
jbygdell, line 201:
```suggestion
	filesURL, _ := url.JoinPath(baseURL + "/metadata/datasets/" + dataset + "/files")
```

The `url.JoinPath` take care of any multiple slashes and it is safe to ignore the error since we do `url.ParceRequestURI` above.

sda_download/sda_download.go
kostas-kou: Right

sda_download/sda_download.go
```diff
@@ -0,0 +1,257 @@
...
+		filePathSplit := strings.Split(filePath, "/")
+		if strings.Contains(filePathSplit[0], "_") {
+			userIDSplit := strings.Split(filePathSplit[0], "_")
+			userSHA := userIDSplit[0]
+			userSHARegex := regexp.MustCompile("^[a-f0-9]{40}$")
+			match := userSHARegex.MatchString(userSHA)
+			if match {
+				filePath = strings.Join(filePathSplit[1:], "/")
+			}
+		}
```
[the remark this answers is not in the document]
kostas-kou: This had to change to replaceall, linter was complaining

sda_download/sda_download.go
kostas-kou: That had to change back because with this way the url.JoinPath was giving the wrong filesURL, it was removing one slash from the dataset which was `https://doi.example/ty009.sfrrss/600.45asasga`. As a result the filesURl was `http://localhost:8080/metadata/datasets/https:/doi.example/ty009.sfrrss/600.45asasga/files` and of course that was not correct

jbygdell: Did you add a test case with the funky doi URL as dataset name so we keep testing for it?

sda_download/sda_download_test.go
```diff
@@ -0,0 +1,176 @@
...
+	assert.Error(suite.T(), err)
+
+	// Test with fileID
+	filepath = "file1id"
+	_, err = getFileIDURL(baseURL, token, dataset, filepath)
+	assert.NoError(suite.T(), err)
+
+	// Testr with bad URL
+	_, err = getFileIDURL("some/url", token, dataset, filepath)
+	assert.Error(suite.T(), err)
```
jbygdell, line 144: Add a test where the dataset name is an URL

kostas-kou: Done

MalinAhlberg (approved):
Looks good! And very nice that you provided an easy-to-run test setup!

jbygdell (dismissed):

MalinAhlberg (approved):

aaperis (approved):
It looks great! Some minor comments regarding documentation :-)

sda_download/sda_download.go
```diff
@@ -0,0 +1,262 @@
...
+
+// Args is a flagset that needs to be exported so that it can be written to the
+// main program help
+var Args = flag.NewFlagSet("sda-download", flag.ExitOnError)
+
+var configPath = Args.String("config", "", "S3 config file to use for downloading.")
+
+var datasetID = Args.String("dataset", "", "Dataset ID for the file to download")
+
+var URL = Args.String("url", "", "The name of the file to download")
```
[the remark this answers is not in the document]
nanjiangshu: I agree with Alex and this should be fixed.

README.md
```diff
@@ -174,9 +174,14 @@ If no config is given by the user, the tool will look for a previous login from
...
 ## Download
 
-The SDA/BP archive enables for downloading files and datasets in a secure manner. That can be achieved using the `sda-cli` tool and the process consists of the following two steps
+The SDA/BP archive enables for downloading files and datasets in a secure manner. That can be achieved using the `sda-cli` tool and and it can be done in two ways:
+- by downloading from a S3 bucket
+- by using the download API
 
-### Create keys
+### Download from S3 bucket
+This process consists of the following two steps: create keys and download file and is explained in the following sections.
```
nanjiangshu, line 182:
```suggestion
This process consists of the following two steps: creating keys and downloading the file. These steps are explained in the following sections.
```

README.md
```diff
@@ -174,9 +174,14 @@ If no config is given by the user, the tool will look for a previous login from
 
 ## Download
 
-The SDA/BP archive enables for downloading files and datasets in a secure manner. That can be achieved using the `sda-cli` tool and the process consists of the following two steps
+The SDA/BP archive enables for downloading files and datasets in a secure manner. That can be achieved using the `sda-cli` tool and and it can be done in two ways:
+- by downloading from a S3 bucket
+- by using the download API
```
nanjiangshu, line 179: Maybe specify which sub commands are related two these two methods?

README.md
```diff
@@ -204,6 +209,19 @@ The tool also allows for selecting a folder where the files will be downloaded,
 ```
 **Note**: If needed, the user can download a selection of files from an available dataset by providing a customized `urls_list.txt` file.
 
+### Download using the download API
+
+The download API allows for downloading files from the archive and it requires the user to have access to the dataset, therefore a configuration file needs to be downloaded before starting the downloading of the files.
```
nanjiangshu, line 214: Will be better to add a cross link to the section `### Download the configuration file`

nanjiangshu (commented):
It works very well when I tested. A general feeling is that the naming of sub commands `download` and `sda-download` does not clearly distinguish methods for downloading. 
I don't have any good suggestions neither. Maybe 
```
s3-download:  by downloading from a S3 bucket
api-download: by using the download API

```

README.md
kostas-kou: done

README.md
kostas-kou: done

README.md
kostas-kou: done

PR #408, opened 2024-07-01
Feature/download dataset

### LINKED ISSUE #387: Download a full dataset with the sda-cli
As a user 
I want to download a dataset that I have been granted access to
so that I get directly my files without interacting with sda-devs.

Description:
Extend the functionality implemented for downloading a file to downloading a whole dataset.
 
A/C:
- giving just `--dataset` flag without filepath arguments downloads the full dataset to the specified out directory

Demo:
- the above functionality

kostas-kou:
This PR closes #387.

**Description**
A new feature in the sda-cli sda-download is implemented in order to download all the files of the dataset (decrypted)
by adding the flag `--dataset`.

**How to test**
Run the setup script from the root folder by typing: `bash .github/integration/setup/setup.sh`. When everything is ready and the containers are running download all the files of the `https://doi.example/ty009.sfrrss/600.45asasga`dataset by running: `./sda-cli sda-download -config testing/s3cmd-download.conf -datasetID https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 -outdir test-dataset --dataset`.
Then check the `test-dataset` folder and 3 files in different paths will be present.


sda_download/sda_download.go
```diff
@@ -106,29 +105,56 @@ func SdaDownload(args []string) error {
...
-		// Check if the file path contains a userID and if it does,
-		// do not keep it in the file path
-		filePathSplit := strings.Split(filePath, "/")
-		if strings.Contains(filePathSplit[0], "_") {
-			_, err := mail.ParseAddress(strings.ReplaceAll(filePathSplit[0], "_", "@"))
-			if err == nil {
-				filePath = strings.Join(filePathSplit[1:], "/")
-			}
+func datasetCase(token string) error {
+	fmt.Println("No files provided, downloading all files in the dataset")
```
nanjiangshu, line 127: This message is a bit misleading, when the command is aimed for downloading the whole dataset, it doesn't matter if files are provided or not. 
Maybe just change it to `Downloading all files in the dataset`

nanjiangshu (commented):
Works well when I tested. Good job @kostas-kou! 

I have some suggestions for the command line syntax. 

1. The flag `-datasetID` is not very good because the use of upper case letters. Firstly, it is not common to use upper case, and secondly, it is easy to make mistakes when typing the command when it has a combination of upper and lower case. I would suggest to change it to `-datasetid` or `-dataset-id`
2. The flag `-dataset` is not very clear to deliver the message that all files will be downloaded for the given dataset ID. I think `-all` will be better. Then the options will become `(--all | [filepath(s)])`, which is more intuitive 
3. When the flag `-dataset` is set, and if one also append some filepaths, the command just went through and download all files in the dataset. I have added another comment in the code that the message `No files provided, downloading all files in the dataset` is a bit misleading, since all files will be downloaded even filepaths are provided. 
We can probably add a validation in the argument parsing so that when `-dataset` is set, error will be reported when filepaths are also provided. That is, either `-dataset` or `filepaths` list can be provided in the command, but not both.

sda_download/sda_download.go
kostas-kou: You are right, I have changed it

kostas-kou:
> Works well when I tested. Good job @kostas-kou!
> 
> I have some suggestions for the command line syntax.
> 
>     1. The flag `-datasetID` is not very good because the use of upper case letters. Firstly, it is not common to use upper case, and secondly, it is easy to make mistakes when typing the command when it has a combination of upper and lower case. I would suggest to change it to `-datasetid` or `-dataset-id`
> 
>     2. The flag `-dataset` is not very clear to deliver the message that all files will be downloaded for the given dataset ID. I think `-all` will be better. Then the options will become `(--all | [filepath(s)])`, which is more intuitive
> 
>     3. When the flag `-dataset` is set, and if one also append some filepaths, the command just went through and download all files in the dataset. I have added another comment in the code that the message `No files provided, downloading all files in the dataset` is a bit misleading, since all files will be downloaded even filepaths are provided.
>        We can probably add a validation in the argument parsing so that when `-dataset` is set, error will be reported when filepaths are also provided. That is, either `-dataset` or `filepaths` list can be provided in the command, but not both.

The first and the third comment are fixed now. For the second one I cannot do anything since the task was saying to add the `--dataset` flag for downloading all the files. We had discussed that when the card was estimated and that was the decision of the team. Of course we can discuss it further if you would prefer a different flag and change it during a refactoring sprint

nanjiangshu:
> The first and the third comment are fixed now. For the second one I cannot do anything since the task was saying to add the `--dataset` flag for downloading all the files. We had discussed that when the card was estimated and that was the decision of the team. Of course we can discuss it further if you would prefer a different flag and change it during a refactoring sprint

I see. Obviously I have missed that there was a decision about the flag name. I'm fine with it for now.

kostas-kou:
There where conflicts after the merged PR, I tried to resolve them on the fly and it failed. I am converting this PR to draft until everything is fixed

nanjiangshu (approved):

pahatz (approved):
It works well now. 
I agree with the `-dataset-id --dataset` being in the same command. It looks a bit weird, but also adds inconsistency with `-dataset` being used on the `htsget` module.
Approving it now, but we should probably discuss this.

PR #417, opened 2024-07-12
Feature/download fileid

### LINKED ISSUE #414: Download file by giving the file id
As a user
I want to provide the file id in sda-download and download the file
so that I do not have to give the whole file path

Description:
This feature is implemented in the sda-download but is poorly implemented.
If a user provide a file id, then the file will be downloaded but not with the proper path
but instead it will just download a file which is named with the file id.

Steps:
- when the download file function (downloadFile) is called for the fileCase then the given filepath is provided as input and it should be changed to the file.Filepath
- Add unit test for the case of downloading file by giving the file id
- Add this case in the tests.sh script

kostas-kou:
This PR closes #414.

**Description**
It was implemented in the sda_download that the user can download files by providing the file id. 
This PR:
- fixes the download path which was wrong when this case was used
- added this case in the integration test
- updated the readme file with info about downloading by providing the file id 

**How to test**
Spin up containers by running:
```bash
```
and then run:
```bash
./sda-cli sda-download -config testing/s3cmd-download.conf -dataset-id https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 -outdir download-fileid urn:neic:001-001
```
and you should see the file in the `download-fileid/main/subfolder/dummy_data` path

### DIFF
--- a/sda_download/sda_download.go
+++ b/sda_download/sda_download.go
@@ -45,10 +45,10 @@ var ArgHelp = `
 		All flagless arguments will be used as sda-download uri.
 	[filepath(s)]
 		The filepath of the file to download.
-    [fileid(s)]
-        The file ID of the file to download.
-    [dirpath]
-        The directory path to download all files recursively.`
+    	[fileid(s)]
+        	The file ID of the file to download.
+    	[dirpath]
+        	The directory path to download all files recursively.`
 
 // Args is a flagset that needs to be exported so that it can be written to the
 // main program help
--- a/sda_download/sda_download.go
+++ b/sda_download/sda_download.go
@@ -25,7 +25,7 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help download` command
 var Usage = `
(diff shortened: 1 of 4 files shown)


sda_download/sda_download.go
```diff
@@ -44,8 +44,9 @@ var ArgHelp = `
 	[uri]
 		All flagless arguments will be used as sda-download uri.
 	[filepath(s)]
-		The filepath of the file to download. If no filepath is provided
-        then the whole dataset will be downloaded.
+		The filepath of the file to download.
+    [fileid(s)]
```
pahatz, line 48: Fix the indentation here

pahatz (approved):
Works great!

sda_download/sda_download.go
pahatz: dirpath below is also off

kostas-kou: strange, I do not see them off locally

kostas-kou: I fixed it on spot because in my case they were aligned

PR #443, opened 2024-08-29
Feature/list dataset show more info

### LINKED ISSUE #433: [list] returns filepaths and dataset size of a dataset
As a user 
I want to list the files in a dataset for which I have access to and the files therein and their sizes using the cli tool
so that I can plan how to proceed with downloading files.

### Description:
Finish the A/C from #388 that were not implemented in #415, i.e. filepaths should be returned instead of filenames so that they can used with 
`./sda-cli download ` 
and dataset size should be returned so that the user knows what to expect if the whole dataset is to be downloaded.

### A/C;
- `sda-cli list --config configfile --url <download.se> --dataset ABC` gives the list of filepaths/fileID,  their size and the dataset size
- It would also be nice to have the fileIDs in a 3rd column
- have size in bytes output as an option `--bytes`

### Example output
```sh
$ sda-cli list [...]
<some metadata>
FileID          Size     Path
ieNgoothoo6e    261kb    dirA/fileA.bam
Vuh8Oochepah    220mb    dirA/fileB.bam
ol5Ir7aeKaiC     64kb    dirA/fileC.bam
en9Voo7Uuch5    230mb    dirA/fileD.bam
Gohxah4yuqu3     46gb    dirA/dirB/fileE.bam
johg4aiGiu0i    362kb    dirA/dirB/fileF.bam
iej1Hoh3eDai    171mb    dirA/dirC/fileG.bam
ahch5iechohT    171mb    dirA/dirC/fileI.bam
eeFooSoh2doo     24gb    dirA/dirC/fileJ.bam
vui9Ahdah5Ru    378tb    dirA/dirC/fileK.bam

```

MalinAhlberg:
This PR closes #433.

**Description**
- `sda-cli list --dataset X` will print the file ID, file size (in human or machine readable form) and file path of each file in dataset X. Also the dataset size will be printed.
- flag `--bytes` is added, defaulting to `false`. If `true`, file sizes are printed in byte format.

**How to test**
From the root folder run the setup script:
```
```
(note to future self: if it does not work, remove all containers and run `docker volume rm -f $(docker volume ls -f "dangling=true")`).
Then run:
```
./sda-cli list -config testing/s3cmd-download.conf -dataset https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 
```
And get
```
FileID               	 Size       	 Path
urn:neic:001-001 	 1.0 MB 	 5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8_elixir-europe.org/main/subfolder/dummy_data.c4gh
urn:neic:001-002 	 1.0 MB 	 5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8_elixir-europe.org/main/subfolder2/dummy_data2.c4gh
urn:neic:001-003 	 1.0 MB 	 5baa61e4c9b93f3f0682250b6cf8331b7ee68fd8_elixir-europe.org/main/subfolder2/random/dummy_data3.c4gh
Dataset size: 3.1 MB
```

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -18,7 +18,7 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help list` command
 var Usage = `
-USAGE: %s list [--config <s3config-file>] [prefix] (-url <uri> --datasets) (-url <uri> --dataset <dataset-id>) [--bytes]
+USAGE: %s list [-config <s3config-file>] [prefix] (-url <uri> --datasets) (-url <uri> --dataset <dataset-id>) [-bytes]
 
 list:
     Lists recursively all files under the user's folder in the Sensitive
--- a/list/list.go
+++ b/list/list.go
@@ -3,11 +3,13 @@ package list
 import (
 	"flag"
 	"fmt"
+	"strconv"
 
 	"strings"
 
 	"github.com/NBISweden/sda-cli/download"
 	"github.com/NBISweden/sda-cli/helpers"
+	"github.com/dustin/go-humanize"
 	"github.com/inhies/go-bytesize"
 )
 
@@ -16,7 +18,7 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help list` command
 var Usage = `
-USAGE: %s list [-config <s3config-file>] [prefix] (-url <uri> --datasets) (-url <uri> --dataset <dataset-id>)
+USAGE: %s list [--config <s3config-file>] [prefix] (-url <uri> --datasets) (-url <uri> --dataset <dataset-id>) [--bytes]
 
 list:
     Lists recursively all files under the user's folder in the Sensitive
@@ -44,6 +46,8 @@ var configPath = Args.String("config", "",
 var URL = Args.String("url", "", "The url of the sda-download server")
 
 var datasets = Args.Bool("datasets", false, "List all datasets in the user's folder.")
+
+var bytesFormat = Args.Bool("bytes", false, "Print file sizes in bytes (not human-readable format).")
 
 var dataset = Args.String("dataset", "", "List all files in the specified dataset.")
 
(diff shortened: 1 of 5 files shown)


jbygdell (approved):

list/list.go
```diff
@@ -16,7 +18,7 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help list` command
 var Usage = `
-USAGE: %s list [-config <s3config-file>] [prefix] (-url <uri> --datasets) (-url <uri> --dataset <dataset-id>)
+USAGE: %s list [--config <s3config-file>] [prefix] (-url <uri> --datasets) (-url <uri> --dataset <dataset-id>) [--bytes]
```
nanjiangshu, line 21: I think for all flags in `sda-cli` (or most of them?), both `--` and `-` work. And in the description of argument, e.g. 

```
Command line arguments:
  -bytes
    	Print file sizes in bytes (not human-readable format).
  -config string
    	S3 config file to use for listing.
  -dataset string
    	List all files in the specified dataset.
  -datasets
    	List all datasets in the user's folder.
  -url string
    	The url of the sda-download server

    [prefix]
        The location/folder of the s3 to list contents.
```
it's always single dash `-` is used. So no need to change to `--config` in this case. Otherwise, we need to make the usage consistent at all places.

nanjiangshu (approved):
Looks good! Just a minor comment about the usage of single dash `-` or double dash `--` in the usage description.

list/list.go
MalinAhlberg: Good catch, I didn't notice that I added that extra dash. Will fix.

MalinAhlberg:
> Looks good! Just a minor comment about the usage of single dash `-` or double dash `--` in the usage description.

Thanks @nanjiangshu ! Fixed by d1a2b85, please resolve when happy.

PR #444, opened 2024-08-29
List datasets more info

### LINKED ISSUE #438: [list] Show more metadata information when showing information about datasets
As a user
I want to know a little bit more about a dataset before listing its files
So I can take better decisions for the next step

### Description
Currently when listing datasets only the ids of the datasets are displayed. Would be good to also display the total size of a dataset and the number of files it contains.

A/C:
output as in below

Example:
```
$ sda-cli list --datasets
Dataset      Files    Size
dataset-01     345     50 Gb
dataset-02      12     22 Gb
dataset-03    9821    2.7 Tb
```

MalinAhlberg:
This PR closes #438 .

**Description**
`sda-cli list -datasets` will print number of files and dataset size:
```
DatasetID                                	 Files 	 Size
https://doi.example/ty009.sfrrss/600.45asasga 	 3 	 3.1 MB
https://doi.example/ty009.sfrrss/Dataset         1 	 1.1 MB
```

**How to test**
From the root folder run the setup script:
```
```
(note to future self: if it does not work, remove all containers and run `docker volume rm -f $(docker volume ls -f "dangling=true")`).
Then run:
```
./sda-cli list -config testing/s3cmd-download.conf -datasets -url http://localhost:8080 
```
And get
```
DatasetID                                	 Files 	 Size
https://doi.example/ty009.sfrrss/600.45asasga 	 3 	 3.1 MB
```

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -139,9 +139,24 @@ func Datasets(token string) error {
 	if err != nil {
 		return err
 	}
+
 	// Loop through the datasets and list them
 	for _, dataset := range datasets {
-		fmt.Printf("%s \n", dataset)
+		files, err := download.GetFilesInfo(*URL, dataset, "", token)
+		if err != nil {
+			return err
+		}
+		// Set rather long minimum column widths, so that header matches the rest of the table
+		fileIDWidth := 40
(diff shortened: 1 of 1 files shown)


jbygdell (approved):

aaperis (approved):
LGTM, I also like the "note to self" on the description :smile:

PR #447, opened 2024-09-12
Upload files by supplying the token as an ENV or on the command line

### LINKED ISSUE #395: [upload/session] Ability to upload files without storing credentials in the filesystem
# User story
As a EFPIA (umbrella term for EU pharmaceutical companies) user,
I want to be able to supply the access token in then environment
so that I can upload files without having to store credentials in the filesystem.

# Technical details
- be able to supply user credentials as env variables (e.g. username, token)
- sda-cli reads credental variables from env as well and gives precedence to env variables

# Acceptance criteria
- [ ] Be able to do an upload to the bigpicture plattform without credentials in the used config file

# Demo
Upload a file to the bigpicture platform using only accesstoken in environment with no credentials in the config file.

# Out of scope
How the config file without credentials appear on the system. It's fine for this issue if the user downloads the config file and then has to edit the file and change their environment manually.

**Additional context**

jbygdell:
This PR closes #395 .

**Description**
This enables uploading of files where the token is not in the config file but supplied on the command line (`-accesToken`) or as an ENV  `ACCESSTOKEN`. 
Token priority order:
1. Using the `-accessToken` flag
2. From the ENV `ACCESSTOKEN`
3. In the config file

**How to test**
* Build the app
* Take a working config file and remove the `access_token` field
* Upload to the inbox by supplying the accesstoken in the two ways available
  * on the command line with the `-accesToken` flag
  * or with the token in the ENV `ACCESSTOKEN`


MalinAhlberg:
Nice! I think it would be even nicer if the README was updated too, also if the precedence of the token was stated there.
Further, if I don't provide any token at all, I get `Error: could not parse token, reason: token contains an invalid number of segments`. Would it make sense to add something more informative, like "No access token provided"..?

upload/upload.go
```diff
@@ -25,7 +25,7 @@ import (
 // Usage text that will be displayed as command line help text when using the
 // `help upload` command
 var Usage = `
-USAGE: %s upload -config <s3config-file> (--encrypt-with-key <public-key-file>) (--force-overwrite) (--force-unencrypted) (-r) [file(s) | folder(s)] (-targetDir <upload-directory>)
+USAGE: %s upload -config <s3config-file> (-accessToken <access-token>) (--encrypt-with-key <public-key-file>) (--force-overwrite) (--force-unencrypted) (-r) [file(s) | folder(s)] (-targetDir <upload-directory>)
```
nanjiangshu, line 28: A bit weird that we have a mixed flag naming style, e.g. `accessToken` and `encrypt-with-key`. Shall we choose just one style?

upload/upload.go
```diff
@@ -63,7 +63,7 @@ var pubKeyPath = Args.String("encrypt-with-key", "",
 		"Only unencrypted data should be provided when this flag is set.",
 )
 
-var accessToken = Args.String("accessToken", "", "Accesstoken to the inbox service. (optional, if it is set in the config file)")
+var accessToken = Args.String("accessToken", "", "Accesstoken to the inbox service.\n(optional, if it is set in the config file or exported as the ENV `ACCESSTOKEN`)")
```
nanjiangshu, line 66:
```suggestion
var accessToken = Args.String("accessToken", "", "Access token to the inbox service.\n(optional, if it is set in the config file or exported as the ENV `ACCESSTOKEN`)")
```

nanjiangshu (commented):
Works great when I tested!
I also have the same feeling as Malin about the error message when no token is provided. Also, would be great to update the README and describe also the order how token is retrieved from the command.

jbygdell:
> Nice! I think it would be even nicer if the README was updated too, also if the precedence of the token was stated there. Further, if I don't provide any token at all, I get `Error: could not parse token, reason: token contains an invalid number of segments`. Would it make sense to add something more informative, like "No access token provided"..?

That error comes directly from the `CheckTokenExpiration` function.

But a check if `any` token is supplied could probably fit somewhere.

MalinAhlberg (approved):
:+1:

pahatz (approved):
looks great!

PR #451, opened 2024-09-19
Decrypt multiple files without aborting 

### LINKED ISSUE #446: Decrypting multiple files with sda-cli impossible when one of the files has been decrypted already
**Describe the bug**
Decrypting multiple files with sda-cli breaks when one of the files has been decrypted already.

**Steps to reproduce**
1. make a folder with several encrypted files
2. run sda-cli to decrypt one of the files in the folder
3. run sda-cli again to decrypt all files in that folder; this will cause the command to stop because one file is decrypted

**Expected behavior**
- [ ] 

**Additional context**

nanjiangshu:
This PR closes #446 

**Description**
When decrypting multiple files, `sda-cli` would previously abort the process if it encountered an existing decrypted file or a missing encrypted file. After the fix, the process will continue in such cases, and warning messages will be printed instead.

**How to test**
Create some files, e.g. file1.txt file2.txt file3.txt, file4.txt and encrypt them with a public key. 
remove some unencrypted files. 
run the command with 
```
sda-cli decrypt -key private-key-file file*.txt.c4gh nonexist.c4gh
```
Those encrypted files without any existing decrypted files will be decrypted and warning messages will be printed.

### DIFF
--- a/decrypt/decrypt.go
+++ b/decrypt/decrypt.go
@@ -84,7 +84,7 @@ func Decrypt(args []string) error {
 	for i, file := range files {
 		switch {
 		case !helpers.FileIsReadable(file.Encrypted):
-			fmt.Fprintf(os.Stderr, "Warning: cannot read input file %s\n", file.Encrypted)
+			fmt.Fprintf(os.Stderr, "Error: cannot read input file %s\n", file.Encrypted)
 		case *forceOverwrite:
 			fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
 			err := decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
--- a/decrypt/decrypt.go
+++ b/decrypt/decrypt.go
@@ -79,41 +79,26 @@ func Decrypt(args []string) error {
 		return err
 	}
 
-	var warnings []string
-	// Check that all the encrypted files exist, and all the unencrypted don't
-	for _, file := range files {
-		// check that the input file exists and is readable
-		if !helpers.FileIsReadable(file.Encrypted) {
-			warnings = append(warnings, fmt.Sprintf("Warning: cannot read input file %s", file.Encrypted))
-
-			continue
-		}
-
-		// check that the output file doesn't exist
-		if helpers.FileExists(file.Unencrypted) && !*forceOverwrite {
-			warnings = append(warnings, fmt.Sprintf("Warning: output file %s already exists", file.Unencrypted))
-		}
-	}
-
-	// Print warnings if any
-	if len(warnings) > 0 {
-		for _, warning := range warnings {
-			fmt.Println(warning)
-		}
-	}
-
 	// decrypt the input files
 	numFiles := len(files)
 	for i, file := range files {
-		if helpers.FileIsReadable(file.Encrypted) && !helpers.FileExists(file.Unencrypted) || *forceOverwrite {
+		switch {
+		case !helpers.FileIsReadable(file.Encrypted):
+			fmt.Fprintf(os.Stderr, "Warning: cannot read input file %s\n", file.Encrypted)
+		case *forceOverwrite:
 			fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
 			err := decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
 			if err != nil {
-				fmt.Printf("Error decrypting file %s: %v\n", file.Encrypted, err)
+				fmt.Fprintf(os.Stderr, "Error decrypting file %s: %v\n", file.Encrypted, err)
 			}
-		} else if helpers.FileExists(file.Unencrypted) {
-			// Skip decrypting if the file already exists and forceOverwrite is not set
-			fmt.Printf("Skipping decryption for file %s as it already exists and forceOverwrite is not enabled\n", file.Unencrypted)
+		case helpers.FileExists(file.Unencrypted):
+			fmt.Fprintf(os.Stderr, "Warning: file %s is already decrypted, skipping\n", file.Unencrypted)
+		default:
+			fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
+			err := decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Error decrypting file %s: %v\n", file.Encrypted, err)
+			}
 		}
 	}
 
--- a/decrypt/decrypt.go
+++ b/decrypt/decrypt.go
@@ -79,26 +79,40 @@ func Decrypt(args []string) error {
 		return err
 	}
 
+	var warnings []string
 	// Check that all the encrypted files exist, and all the unencrypted don't
 	for _, file := range files {
 		// check that the input file exists and is readable
 		if !helpers.FileIsReadable(file.Encrypted) {
-			return fmt.Errorf("cannot read input file %s", file.Encrypted)
+			warnings = append(warnings, fmt.Sprintf("Warning: cannot read input file %s", file.Encrypted))
+			continue
 		}
 
 		// check that the output file doesn't exist
 		if helpers.FileExists(file.Unencrypted) && !*forceOverwrite {
-			return fmt.Errorf("outfile %s already exists", file.Unencrypted)
+			warnings = append(warnings, fmt.Sprintf("Warning: output file %s already exists", file.Unencrypted))
+		}
+	}
+
+	// Print warnings if any
+	if len(warnings) > 0 {
+		for _, warning := range warnings {
+			fmt.Println(warning)
 		}
 	}
 
 	// decrypt the input files
 	numFiles := len(files)
 	for i, file := range files {
-		fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
-		err = decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
-		if err != nil {
-			return err
+		if helpers.FileIsReadable(file.Encrypted) && !helpers.FileExists(file.Unencrypted) || *forceOverwrite {
+			fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
+			err := decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
+			if err != nil {
+				fmt.Printf("Error decrypting file %s: %v\n", file.Encrypted, err)
+			}
+		} else if helpers.FileExists(file.Unencrypted) {
+			// Skip decrypting if the file already exists and forceOverwrite is not set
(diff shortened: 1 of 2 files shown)


decrypt/decrypt.go
```diff
@@ -79,26 +79,41 @@ func Decrypt(args []string) error {
...
-			return err
+		if helpers.FileIsReadable(file.Encrypted) && !helpers.FileExists(file.Unencrypted) || *forceOverwrite {
+			fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
+			err := decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
+			if err != nil {
+				fmt.Printf("Error decrypting file %s: %v\n", file.Encrypted, err)
+			}
+		} else if helpers.FileExists(file.Unencrypted) {
+			// Skip decrypting if the file already exists and forceOverwrite is not set
+			fmt.Printf("Skipping decryption for file %s as it already exists and forceOverwrite is not enabled\n", file.Unencrypted)
```
jbygdell, line 116:
```suggestion
		switch {
		case !helpers.FileIsReadable(file.Encrypted):
			fmt.Fprintf(os.Stderr, "Warning: cannot read input file %s\n", file.Encrypted)
		case *forceOverwrite:
			fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
			err := decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
			if err != nil {
				fmt.Fprintf(os.Stderr, "Error decrypting file %s: %v\n", file.Encrypted, err)
			}
		case helpers.FileExists(file.Unencrypted):
			fmt.Fprintf(os.Stderr, "Warning: file %s is already decrypted, skipping\n", file.Unencrypted)
		default:
			fmt.Printf("Decrypting file %v/%v: %s\n", i+1, numFiles, file.Encrypted)
			err := decryptFile(file.Encrypted, file.Unencrypted, *privateKey)
			if err != nil {
				fmt.Fprintf(os.Stderr, "Error decrypting file %s: %v\n", file.Encrypted, err)
			}
```

If you use a switch case like this not only will you cleanup the nested if-else-if block but you will also be able to remove the code block above where the files are checked to no real use.

decrypt/decrypt.go
```diff
@@ -79,26 +79,41 @@ func Decrypt(args []string) error {
 		return err
 	}
 
+	var warnings []string
 	// Check that all the encrypted files exist, and all the unencrypted don't
 	for _, file := range files {
 		// check that the input file exists and is readable
 		if !helpers.FileIsReadable(file.Encrypted) {
-			return fmt.Errorf("cannot read input file %s", file.Encrypted)
+			warnings = append(warnings, fmt.Sprintf("Warning: cannot read input file %s", file.Encrypted))
```
jbygdell, line 87: This is a bit of a semantic problem, should be just continue print warnings if a file is not readable? Either that the file doesn't exists or that there are permission problems.

jbygdell (changes requested):

decrypt/decrypt.go
nanjiangshu: Checking whether a file is readable without determining the cause, whether due to its absence or permission issues, is used elsewhere in the code. If this is needed, I think it would be better to address this in a new refactoring issue.

jbygdell: That might be but with this PR the semantics are changed, abort -> warning so the actual root cause should be shown to the user (file missing or file not readable).

We should add an issue about this since the changes will affect other parts of the code.

decrypt/decrypt_test.go
```diff
@@ -156,3 +156,81 @@ func (suite *DecryptTests) TestDecrypt() {
 	assert.NoError(suite.T(), err, "decrypt failed unexpectedly")
 	os.Args = nil
 }
+
+func (suite *DecryptTests) TestDecryptMultipleFiles() {
```
jbygdell, line 160: Either expand this test or create another one to cover the following case:
* ForceOverwright = true
* overwrite an empty file and verify that the file has new contents

nanjiangshu: Added in the commit 42c404e

The failure of the s3 integration test is caused by the PR https://github.com/neicnordic/sensitive-data-archive/pull/1046. It seems the image is not successfully created for the tag v0.4.0

MalinAhlberg (approved):
Nice!

decrypt/decrypt_test.go
```diff
@@ -156,3 +156,119 @@ func (suite *DecryptTests) TestDecrypt() {
...
+		decryptedFile := filepath.Join(suite.tempDir, filepath.Base(fileName))
+
+		// Clear the file content by truncating the file
+		err := os.WriteFile(decryptedFile, []byte{}, 0600)
+		assert.NoError(suite.T(), err, "failed to empty the file content")
+	}
+
+	// Now, run decryption again with --force-overwrite enabled
+	os.Args = []string{"decrypt", "-key", fmt.Sprintf("%s.sec.pem", testKeyFile),
+		"--force-overwrite", encryptedFiles[0], encryptedFiles[1], encryptedFiles[2]}
```
jbygdell, line 246: Don't beak a set like this, supply it as one line or with one object on each line. 

```suggestion
	os.Args = []string{"decrypt", "-key", fmt.Sprintf("%s.sec.pem", testKeyFile), "--force-overwrite", encryptedFiles[0], encryptedFiles[1], encryptedFiles[2]}
```

```
	os.Args = []string{
		"decrypt",
		"-key",
		fmt.Sprintf("%s.sec.pem", testKeyFile),
		"--force-overwrite",
		encryptedFiles[0],
		encryptedFiles[1],
		encryptedFiles[2]
	}
```

jbygdell (approved):
One small comment, but other than that it looks fine

aaperis:
There was a comment at some point (I believe by @yohell ) that the info about the skipped files will be buried will be buried in the output logs. For example here:
```
./sda-cli decrypt -key alex.sec.pem b/* 
✔ Enter password to unlock private key: █
Decrypting file 1/6: b/file-list-1.c4gh
Warning: file b/file-list-2 is already decrypted, skipping
Warning: file b/file-list-2 is already decrypted, skipping
Decrypting file 4/6: b/file-list-3.c4gh
Warning: file b/file-list-4 is already decrypted, skipping
Warning: file b/file-list-4 is already decrypted, skipping
```
imagine having 1000 files. I think it would be preferable to have something like the output of `encrypt`  where the user get a note at the end of the output regarding what files got skipped:
```
./sda-cli encrypt -key alex.pub.pem a/* -continue
ERRO[0000] Skipping input file a/file-list-4.c4gh. Reason: input file a/file-list-4.c4gh is already encrypted(.c4gh). 
ERRO[0000] Skipping input file a/file-list-4. Reason: outfile a/file-list-4.c4gh already exists. 
ERRO[0000] Skipping input file a/file-list-2.c4gh. Reason: input file a/file-list-2.c4gh is already encrypted(.c4gh). 
ERRO[0000] Skipping input file a/file-list-2. Reason: outfile a/file-list-2.c4gh already exists. 
ERRO[0000] (4/6) files skipped
```

Although, the output from `encrypt` needs to be changed to non `log.Error` messages.

aaperis:
> > imagine having 1000 files. I think it would be preferable to have something like the output of `encrypt` where the user gets a note at the end of the output regarding what files got skipped:
> 
> The decypt returns output to different channels depending on if it is a notification or Error/Warning. So you can do something like this it that case:
> 
> ```
> sda-cli decrypt -key c4gh.private files/* 1>stdout.log 2>stderr.log
> ```

that does the job, but not as nicely as `encrypt` and I would expect that both commands behave similarly.

aaperis:
I also noticed that the same Warnings are printed twice:
```
./sda-cli decrypt -key alex.sec.pem b/* 
✔ Enter password to unlock private key: 
Decrypting file 1/6: b/file-list-1.c4gh
Warning: file b/file-list-2 is already decrypted, skipping
Warning: file b/file-list-2 is already decrypted, skipping
Decrypting file 4/6: b/file-list-3.c4gh
Warning: file b/file-list-4 is already decrypted, skipping
Warning: file b/file-list-4 is already decrypted, skipping
```

aaperis (changes requested):
Nice work overall but there are some things that imo should be fixed so that output is more consistent, see comments :-)

nanjiangshu:
> Nice work overall but there are some things that imo should be fixed so that output is more consistent, see comments :-)

Thanks for the good suggestion @aaperis , do you want me to modify the code so that the warning messages be printed at the end or we fix it in future issues that make SDA-CLI more coherent?

In addition, I can not see the duplicated warning message when I tried, can you give me some instructions about how to reproduce?

aaperis:
> Thanks for the good suggestion @aaperis , do you want me to modify the code so that the warning messages be printed at the end or we fix it in future issues that make SDA-CLI more coherent?

My thinking was that this should be done now that the PR is still open but if you feel otherwise make an issue about it so that we don't forget it in the upcoming epic :smile: .
 
> In addition, I can not see the duplicated warning message when I tried, can you give me some instructions about how to reproduce?

That's strange, all I did was to have a folder `b` with some encrypted and unencrypted files and run the command that I posted in the above comment. Could be a linux/mac thing?

nanjiangshu:
> > Nice work overall but there are some things that imo should be fixed so that output is more consistent, see comments :-)
> 
> Thanks for the good suggestion @aaperis , do you want me to modify the code so that the warning messages be printed at the end or we fix it in future issues that make SDA-CLI more coherent?
> 
> In addition, I can not see the duplicated warning message when I tried, can you give me some instructions about how to reproduce?

nanjiangshu:
> > Thanks for the good suggestion @aaperis , do you want me to modify the code so that the warning messages be printed at the end or we fix it in future issues that make SDA-CLI more coherent?
> 
> My thinking was that this should be done now that the PR is still open but if you feel otherwise make an issue about it so that we don't forget it in the upcoming epic 😄 .
> 

@aaperis, I updated the description of the issue https://github.com/NBISweden/sda-cli/issues/455 to mention that the order how the warning/error messages should be printed should be solved there. 
I think it is better to solve this problem in that issue since we may find a better coherent way to output the warning/error message.


> > In addition, I can not see the duplicated warning message when I tried, can you give me some instructions about how to reproduce?
> 
> That's strange, all I did was to have a folder `b` with some encrypted and unencrypted files and run the command that I posted in the above comment. Could be a linux/mac thing? or something weird with my setup which I cannot identify..

I tried with both Linux and Mac but could not reproduce the duplicated output. May someone else test to verify it?

aaperis:
> @aaperis, I updated the description of the issue #455 to mention that the order how the warning/error messages should be printed should be solved there. I think it is better to solve this problem in that issue since we may find a better coherent way to output the warning/error message.

works for me

> I tried with both Linux and Mac but could not reproduce the duplicated output. May someone else test to verify it?

seems we'll have to wait and see if anyone reports this weird issue in the future.

Approve :smile:

aaperis (approved):

PR #458, opened 2024-10-16
config  file provided before command

### LINKED ISSUE #442: [session] Config file should be an option for the base command
As a user of the tool
I want options that apply settings for all subcommands to come before subcommands
So that it's easier to understand what is what

# Description

It's weird that the more "global" options are specified for each subcommand instead of globally.

`sda-cli -config <config-file> list --dataset <dataset>`

# A/C
* it's possible to specify the config file as an option to the base command instead of the subcommands
* add tests

# Demo
2 minutes.

pahatz:
This PR closes #442 .

**Description**
The user should be able to provide the config file location argument *ONLY* before the subcommand of the cli. 

**How to test**
For test coverage, some of the existing tests have been altered to follow the new allowed order of providing the config, i.e. before the subcommand.
To manually test, one should test individually the cli commands that are affected. Those are the ones that are using the config as an argument(upload, list htsget, download).

### DIFF
--- a/htsget/htsget_test.go
+++ b/htsget/htsget_test.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"path/filepath"
 	"runtime"
 	"testing"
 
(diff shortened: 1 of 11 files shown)


jbygdell (changes requested):
There shuld be no IFs here, the config must be set before any action i.e. `sda-cli -config cli.conf ACTION -FLAGS` should be the only functioning command.

Config can be as a file or as ENV flags.

`Usage` for all commands needs to be updated to match the new setup.

MalinAhlberg (commented):
Good work, @pahatz! Works well as far as I can tell!
I agree that it is a bit confusing that you can provide two configs, though... unless there's a use case I'm missing, it's probably cleaner to only allow one.

The README also needs some updates (apart from the `Usage` sections in the go files, that Jocke mentioned).

pahatz:
I changed the behavior to work only for the config to be provided before the command.
Updated the tests, readme and Usage references.

jbygdell (changes requested):

htsget/htsget_test.go
```diff
@@ -58,8 +58,8 @@ access_token = eyJ0eXAiOiJKV1QiLCJqa3UiOiJodHRwczovL29pZGM6ODA4MC9qd2siLCJhbGciO
 	if err != nil {
 		panic(err)
 	}
-	os.Args = []string{"htsget", "-config", tmpDir + "s3cmd_test.conf", "-dataset", "DATASET0001", "-filename", "htsnexus_test_NA12878", "-host", "somehost", "-pubkey", "somekey"}
-	err = Htsget(os.Args)
+	os.Args = []string{"htsget", "-dataset", "DATASET0001", "-filename", "htsnexus_test_NA12878", "-host", "somehost", "-pubkey", "somekey"}
+	err = Htsget(os.Args, tmpDir+"s3cmd_test.conf")
```
nanjiangshu, line 62: Although it is the test code, I think it is better to use a library to concatenate path instead of just using `+`.
E.g.

```golang
err = Htsget(os.Args, filepath.Join(tmpDir, "s3cmd_test.conf"))
```
So as for the other places

nanjiangshu (commented):
I have just added a comment regarding path concatenation, otherwise, it looks great!

nanjiangshu (commented):
Looks good to me. It's weird that the integration test failed even after I reran it. Maybe a rebase to the main will fix the problem.

pahatz:
> Looks good to me. It's weird that the integration test failed even after I reran it. Maybe a rebase to the main will fix the problem.

v0.3.159 appears as the latest tag but the related sda images are missing.

jbygdell:
> > Looks good to me. It's weird that the integration test failed even after I reran it. Maybe a rebase to the main will fix the problem.
> 
> `v0.3.159` appears as the latest tag but the related sda images are missing.

v0.3.158 is the latest released version

main.go
```diff
@@ -98,7 +99,18 @@ func ParseArgs() (string, []string) {
...
+
+	// If config comes before the command, we remove it from the
+	// arguments and set the global config path.
+	if len(os.Args) > 1 && (os.Args[1] == "--config" || os.Args[1] == "-config") {
+		if len(os.Args) < 3 {
+			fmt.Fprintf(os.Stderr, "Error: --config requires an argument\n")
+			os.Exit(1)
+		}
+		configPath = os.Args[2]
+		os.Args = append(os.Args[:1], os.Args[3:]...)
```
jbygdell, line 113:
```suggestion
	case "--config", "-config":
		if len(os.Args) < 3 {
			fmt.Fprintf(os.Stderr, "Error: --config requires an argument\n")
			os.Exit(1)
		}
		configPath = os.Args[2]
		os.Args = append(os.Args[:1], os.Args[3:]...)
```

jbygdell (approved):

MalinAhlberg (approved):

PR #463, opened 2024-10-30
Check for new releases

### LINKED ISSUE #449: [version] Check if newer version is available
As a user
I want to know if I am using the latest version
so that to fetch the latest if I have an older one.

**Description**:
Should be displayed the version of the sda-cli, then checking if a newer version is available when the version command is used. In case of no internet connection the user should be notified as well.

A/C:
- tests

jbygdell:
This PR closes #449 .

**Description**
When running `sda-cli version` the app will check the latest release page on GitHub and match that against the version tag it has. If a newer version is available it will print the following message:

```
Newer version if sda-cli is avaiable, v0.1.3
Published: 2024-09-19 09:23:33
Download it from here: https://github.com/NBISweden/sda-cli/releases/tag/v0.1.3
```

**How to test**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.55.5
 	github.com/dustin/go-humanize v1.0.1
 	github.com/golang-jwt/jwt v3.2.2+incompatible
+	github.com/hashicorp/go-version v1.7.0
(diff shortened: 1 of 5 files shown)


pahatz (approved):
Works great! That library also looks like it might come in handy if we ever expand versioning.
I only pointed out some typos.

aaperis (approved):

PR #464, opened 2024-10-30
Don't write the public key to disk

### LINKED ISSUE #291: [session] don't write the public key to disk
It should be possible to use the public key stored in the token without first having to write it back to disk.

Current implementation of sda-cli when it reads the c4gh pub key from the `/info` endpoint from `sda-auth` it writes it on disk again. This should not be the case.

A/C:
- don't write the key to disk  again once it is in memory

jbygdell:
This PR closes #291.

**Description**
When reading the public key from auths info endpoint we don't need to write the data to disk and then read the public key from the file before it is used for encryption.

**How to test**

### DIFF
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -106,9 +106,12 @@ func Encrypt(args []string) error {
 		if err != nil {
 			return err
 		}
-		// key from session file found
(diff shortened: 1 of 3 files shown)


aaperis (approved):
Nice!

pahatz (approved):

PR #484, opened 2024-12-17
split integration tests

### LINKED ISSUE #403: [tests] split the integration test file
As a developer, I want the integration test file to be split into smaller files, so that they are easier to manage and maintain for future development.

**Description**
Ideally, different sub commands would be placed in a separate bash script.

**Acceptance criteria**
- [ ] Tests that are related or have similar purposes are consolidated within the same test file 

**Additional context**

MalinAhlberg:
This PR closes #403 .

**Description**
The file with integration tests is split into:
```
10_encrypt_decrypt.sh
20_upload.sh
30_list.sh
40_download.sh 
```

**How to test**
Make sure the tests in github actions runs, and that you can run all tests locally, eg:
```
.github/integration/setup/setup.sh  # some manual intervention might be needed here, at least for me, but no changes from before
.github/integration/tests/10_upload.sh
.github/integration/tests/20_encrypt_decrypt.sh
.github/integration/tests/30_list.sh
.github/integration/tests/40_download.sh
```


.github/integration/tests/10_encrypt_decrypt.sh
```diff
@@ -0,0 +1,135 @@
...
+    C4GH_PASSWORD="" ./sda-cli decrypt -key $key.sec.pem data_file_keys.c4gh
+    if [ -f data_file_keys ]; then
+        echo "Decrypted data file"
+    else
+        echo "Failed to decrypt data file with $key"
+        exit 1
+    fi
+done
+echo "Could decrypt with both keys from concatenated key"
+rm data_file_keys.c4gh
```
jbygdell, line 114: Technically this is three different keys

.github/integration/tests/20_upload.sh
```diff
@@ -0,0 +1,115 @@
...
+output=$(./sda-cli -config testing/s3cmd.conf upload --force-overwrite data_file.c4gh)
+
+
+# Test recursive folder upload
+for k in data_file.c4gh data_file1.c4gh
+do
+    # Upload and check file
+    ./sda-cli -config testing/s3cmd.conf upload --force-overwrite "data_files_enc/$k"
+    check_uploaded_file "test/$user/$k" "$k"
+done
```
jbygdell, line 51: This is not a recursive upload as per comment.

jbygdell (approved):
Overall good work.
A bit on the verbose side for my taste, some miss match on comments vs. what's happening.

Instead of having to use `--force-overwrite` when it is not part of the test maybe a bucket cleanup operation would be in place.

MalinAhlberg:
> Overall good work. A bit on the verbose side for my taste, some miss match on comments vs. what's happening.
> 
> Instead of having to use `--force-overwrite` when it is not part of the test maybe a bucket cleanup operation would be in place.

Thanks! I tried not to change the tests (or their comments) much, only split the tests into different files. But I agree that some comments should be updated. Will fix what you suggested and look over the comments in general.

.github/integration/tests/20_upload.sh
MalinAhlberg: Fixed in 38a8ef7.

.github/integration/tests/10_encrypt_decrypt.sh
MalinAhlberg: Fixed in 38a8ef7.

.github/integration/tests/10_encrypt_decrypt.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+set -e
+
+# Function checking that a file was encrypted
```
nanjiangshu, line 4:
```suggestion
# Function checking that file(s) were encrypted
```

.github/integration/tests/10_encrypt_decrypt.sh
```diff
@@ -0,0 +1,125 @@
...
+function check_encypted_file() {
```
nanjiangshu, line 5: This function is not only checking "one file" but checking all files separated by white spaces provided in the first argument. 
Note sure if this is best. I would expect the function is called as 
```
check_encrypted_file file1 file2
```
Instead of 
```
check_encrypted_file "file1 file2"
```

Suggested changes:
1. Update the comment line, indicating that it tries to checking all files provided 
2. Fixing the typo of the function name, i.e. `check_encypted_file` to `check_encrypted_file`
3. Probably change the function to 
e.g. 
```sh
function check_encrypted_file() {
    for k in "$@"; do
        output=$(head -c 8 "$k")

        if [[ "$output" == "crypt4gh" ]]; then
            echo "Encrypted data file: $k"
        else
            echo "Failed to encrypt file: $k"
            exit 1
        fi
    done
}
```
so that the arguments of a list of files are provided in a more natural way.

.github/integration/tests/10_encrypt_decrypt.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# Create another couple of key-pairs
+for c in 1 2
+do
+    if ( yes "" | ./sda-cli createKey sda_key$c ) ; then
+        echo "Created key pair for encryption"
+    else
+        echo "Failed to create key pair for encryption"
+        exit 1
+fi
```
nanjiangshu, line 58:
```suggestion
    fi
```

.github/integration/tests/20_upload.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
+set -e
+# inferred from access_key in testing/s3cmd.conf
+user=test_dummy.org
+
+
+# Function checking that a file was encrypted
+function check_encypted_file() {
```
nanjiangshu, line 8: This function is redefined and not sure if it will be duplicated more in the future. Maybe it is better to define it in a separate function call e.g. `common_func.sh` and then source it from there.

nanjiangshu (commented):
Looks great! Now similar integration tests are separated nicely. I left a few comments for potential improvements.

MalinAhlberg:
Thanks, @nanjiangshu! Those are very valid points, makes sense to include them in this PR. I refactored as you suggested  in 28fb827 (also kudos to @kusalananda for some more refactoring improvements) and rebased on main.

kostas-kou (approved):

.github/integration/tests/20_upload.sh
MalinAhlberg: Fixed

.github/integration/tests/10_encrypt_decrypt.sh
MalinAhlberg: Fixed

.github/integration/tests/10_encrypt_decrypt.sh
MalinAhlberg: Fixed

.github/integration/tests/10_encrypt_decrypt.sh
MalinAhlberg: Fixed

PR #498, opened 2025-01-23
More human friendly token expiry date messages

### LINKED ISSUE #437: [session] Show token expiry time in a more Human friendly format
As a user
I don't want to do complex math to figure out when my token expires
So I can spend my precious, precious brain cycles on more important things.

### Description

Currently token expiry is shown using HHH:MM:SS format, which is complicated to understand, especially when number of hours is more than 48, and also seconds is too fine grained.

Suggestion, following output:
* If more than a day: "on YYYY-MM-DD "
* If less than a day "in 12 hours and 5 minutes (at 23:04)"
* If less than two hours "WARNING, token expires in only 1 hour and 43 minutes"

### Demo
Demo the expiry time using three different tokens (expected demo time, 2 minutes).

aaperis:
This PR closes #437. Now token expiry messages to the user follow the output pattern that is listed in the issue description. 

**How to test**
Tests have been added to the go testsuite, so
```
go test ./...
```
should be enough.

Note: The `download integration` tests fail because of the latest change where `sda-download` serves only `/s3` whereas `sda-cli` looks for `s3-encrypted`. This is irrelevant to this PR though (see temporary fix #501)

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -398,26 +398,26 @@ func CheckTokenExpiration(accessToken string) error {
 	}
 
 	switch untilExp := time.Until(expiration); {
-	case untilExp < 0:
+	case untilExp <= 0:
 		return fmt.Errorf("the provided access token has expired, please renew it")
-	case untilExp >= 0 && untilExp < 2*time.Hour:
+	case untilExp > 0 && untilExp < 2*time.Hour:
 		fmt.Fprintf(
 			os.Stderr,
-			"WARNING! The provided access token expires in only %d hours and %d minutes.\n",
+			"WARNING! The provided access token expires in only %d hour and %d minutes.\n",
 			int(untilExp.Hours()), int(untilExp.Minutes())-(int(untilExp.Hours())*60),
 		)
 		fmt.Fprintln(os.Stderr, "Consider renewing the token.")
 	case untilExp >= 2*time.Hour && untilExp < 24*time.Hour:
 		fmt.Fprintf(
 			os.Stderr,
-			"The provided access token expires in %d hours and %d minutes (at %d:%d).\n",
-			int(untilExp.Hours()), int(untilExp.Minutes())-(int(untilExp.Hours())*60), expiration.Hour(), expiration.Minute(),
+			"The provided access token expires in %d hours and %d minutes.\n",
+			int(untilExp.Hours()), int(untilExp.Minutes())-(int(untilExp.Hours())*60),
 		)
 		fmt.Fprintln(os.Stderr, "Consider renewing the token.")
 	default:
 		fmt.Fprintf(
 			os.Stderr,
-			"The provided access token expires on %s.\n", expiration.Format("2006-01-02"),
+			"The provided access token expires on %s.\n", expiration.Format(time.RFC1123),
 		)
 	}
 
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -398,27 +398,27 @@ func CheckTokenExpiration(accessToken string) error {
 	}
 
 	switch untilExp := time.Until(expiration); {
-	case untilExp <= 0:
-		return fmt.Errorf("the provided access token has expired, please renew it")
-	case untilExp > 0 && untilExp < 2*time.Hour:
+	case untilExp >= 1*time.Minute && untilExp < 2*time.Hour:
 		fmt.Fprintf(
 			os.Stderr,
 			"WARNING! The provided access token expires in only %d hour and %d minutes.\n",
 			int(untilExp.Hours()), int(untilExp.Minutes())-(int(untilExp.Hours())*60),
 		)
 		fmt.Fprintln(os.Stderr, "Consider renewing the token.")
-	case untilExp >= 2*time.Hour && untilExp < 24*time.Hour:
+	case untilExp >= 2*time.Hour && untilExp <= 24*time.Hour:
 		fmt.Fprintf(
 			os.Stderr,
 			"The provided access token expires in %d hours and %d minutes.\n",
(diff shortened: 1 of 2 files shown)


helpers/helpers.go
```diff
@@ -400,18 +400,24 @@ func CheckTokenExpiration(accessToken string) error {
 	switch untilExp := time.Until(expiration); {
 	case untilExp < 0:
 		return fmt.Errorf("the provided access token has expired, please renew it")
-	case untilExp > 0 && untilExp < 24*time.Hour:
-		fmt.Fprintln(
+	case untilExp >= 0 && untilExp < 2*time.Hour:
```
jbygdell, line 403:
```suggestion
	case untilExp >= 1*time.Minute && untilExp < 2*time.Hour:
```

If time remaining is 0 the token is essentially expired.

helpers/helpers.go
```diff
@@ -400,18 +400,24 @@ func CheckTokenExpiration(accessToken string) error {
...
-			time.Until(expiration).Truncate(time.Second),
+			"WARNING! The provided access token expires in only %d hours and %d minutes.\n",
+			int(untilExp.Hours()), int(untilExp.Minutes())-(int(untilExp.Hours())*60),
+		)
+		fmt.Fprintln(os.Stderr, "Consider renewing the token.")
+	case untilExp >= 2*time.Hour && untilExp < 24*time.Hour:
+		fmt.Fprintf(
+			os.Stderr,
+			"The provided access token expires in %d hours and %d minutes (at %d:%d).\n",
+			int(untilExp.Hours()), int(untilExp.Minutes())-(int(untilExp.Hours())*60), expiration.Hour(), expiration.Minute(),
```
jbygdell, line 414: Having the time remaining and the timestamp here feels a bit off. It's better to be consistent and only provide the remaining time.

helpers/helpers.go
```diff
@@ -400,18 +400,24 @@ func CheckTokenExpiration(accessToken string) error {
...
 		)
 		fmt.Fprintln(os.Stderr, "Consider renewing the token.")
 	default:
-		fmt.Fprintln(
+		fmt.Fprintf(
 			os.Stderr,
-			"The provided access token expires in",
-			time.Until(expiration).Truncate(time.Second),
+			"The provided access token expires on %s.\n", expiration.Format("2006-01-02"),
```
jbygdell, line 420: This is the same as using 
```
expiration.Format(time.DateOnly)
```

In this case it might be better to be a bit more verbose so that the user knows if the token will expire outside of working hours.
```suggestion
			"The provided access token expires on %s.\n", expiration.Format(time.RFC1123),
```

This will then print the following message:
```
The provided access token expires on Mon, 27 Jan 2025 08:00:22 CET
```

jbygdell (commented):
While you ar at it consider changing the default response of the switch to handle the case of an expired token.
The logical cases would then be:
1-120 min, 2-24 hours  and > 24 hours.

helpers/helpers.go
aaperis: good point, done

helpers/helpers.go
aaperis: agreed,looks neater this way :smile:

aaperis:
> While you ar at it consider changing the default response of the switch to handle the case of an expired token. The logical cases would then be: 1-120 min, 2-24 hours and > 24 hours.

Makes, sense. Done.

helpers/helpers.go
```diff
@@ -398,21 +398,27 @@ func CheckTokenExpiration(accessToken string) error {
...
-	case untilExp < 0:
-		return fmt.Errorf("the provided access token has expired, please renew it")
-	case untilExp > 0 && untilExp < 24*time.Hour:
-		fmt.Fprintln(
+	case untilExp >= 1*time.Minute && untilExp < 2*time.Hour:
+		fmt.Fprintf(
 			os.Stderr,
-			"The provided access token expires in",
-			time.Until(expiration).Truncate(time.Second),
+			"WARNING! The provided access token expires in only %d hour and %d minutes.\n",
```
nanjiangshu, line 404: Maybe a bit over engineering, bit it's a bit weird to see 
```
The provided access token expires in only 0 hour and 29 minutes.
```
when the token is valid for <30 min

I would expect 
```
 The provided access token expires in only 29 minutes.
```

nanjiangshu (commented):
Looks great! I l left a minor comment for improvement but I'm OK to ignore it as well.

helpers/helpers.go
aaperis: It was quick to fix and I agree it looks better this way.

helpers/helpers.go
```diff
@@ -398,21 +398,31 @@ func CheckTokenExpiration(accessToken string) error {
...
-	case untilExp < 0:
-		return fmt.Errorf("the provided access token has expired, please renew it")
-	case untilExp > 0 && untilExp < 24*time.Hour:
-		fmt.Fprintln(
+	case untilExp >= 1*time.Minute && untilExp < 2*time.Hour:
+		hourString := ""
+		if int(untilExp.Hours()) != 0 {
+			hourString = "1 hour and "
+		}
+		fmt.Fprintf(
```
jbygdell, line 406: I'm not a fan of this style of coding.
This could be handled by another case in the switch:

```suggestion
	case untilExp >= 1*time.Minute && untilExp <= 1*time.Hour:
		fmt.Fprintf(
			os.Stderr,
			"WARNING! The provided access token expires in only %d minutes.\n", 
			int(untilExp.Minutes())-(int(untilExp.Hours())*60),
		)

	case untilExp > 1*time.Hour && untilExp < 2*time.Hour:
```

aaperis: Probably I put too much sauce here. Changed it to another case as suggested.

nanjiangshu (approved):

PR #506, opened 2025-02-14
Remove usage of s3-encrypted

### LINKED ISSUE #497: Handle endpoint renaming
For the new version of sda-download the endpoint /s3-encrypted has been renamed to /s3.
This is only about updating the endpoint that sda-cli expects.
This needs to happen before the maintenance window in March 2025

- Remember to undo commit cf2298453e1fbe67e0652d50f8ef41d91eab1a9e when this is done so that the tests use the latest version of sda-download again.
- make a release for the sda-cli

MalinAhlberg:
This PR closes #497 .

**Description**
-  Usage and references to `/s3-encrypted` are replaced with `/s3`.
- The latest tag will be used for the image of download.
- The endpoint `/files` will not be used anymore, as it is for encrypted data
- Support for removing the user name part of the filepath is improved (but can be removed when we merge [`feature/do-not-leak-user-ids`](https://github.com/neicnordic/sensitive-data-archive/tree/feature/do-not-leak-user-ids) )
- Encrypted files are expected
- Tests are updated 

**How to test**
Use `sda-cli` to dowload files.

**To do**
- [ ] Make a release to v0.1.4

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -196,8 +196,12 @@ func datasetCase(token string) error {
 	// Loop through the files and download them
 	for _, file := range files {
 		// Download URL for the file
-		fileURL := *URL + "/files/" + file.FileID
-		err = downloadFile(fileURL, token, "", file.FilePath)
+		fileName := AnonymizeFilepath(file.FilePath)
+		fileURL := *URL + "/s3/" + file.DatasetID + "/" + fileName
+		if err != nil {
+			return err
+		}
+		err = downloadFile(fileURL, token, pubKeyBase64, file.FilePath)
 		if err != nil {
 			return err
 		}
@@ -236,8 +240,9 @@ func recursiveCase(token string) error {
 		for _, file := range files {
 			if strings.Contains(file.FilePath, dirPath) {
 				pathExists = true
-				fileURL := *URL + "/files/" + file.FileID
-				err = downloadFile(fileURL, token, "", file.FilePath)
+				fileName := AnonymizeFilepath(file.FilePath)
+				fileURL := *URL + "/s3/" + file.DatasetID + "/" + fileName
+				err = downloadFile(fileURL, token, pubKeyBase64, file.FilePath)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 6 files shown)


download/download.go
```diff
@@ -510,3 +507,30 @@ func GetURLsFile(urlsFilePath string) (urlsList []string, err error) {
 
 	return urlsList, scanner.Err()
 }
+
+func AnonymizeFilepath(filePath string) string {
```
nanjiangshu, line 511: Nice refactoring! If this function is supposed to be used by other modules as well, I think it is better to be placed in `helpers.go`. 
And would be even better to add a comment to explain that it is trying to remove email like userID from the filepath. Maybe add a unit text as well?

download/download.go
```diff
@@ -510,3 +507,30 @@ func GetURLsFile(urlsFilePath string) (urlsList []string, err error) {
...
+		_, err := mail.ParseAddress(strings.ReplaceAll(filePathSplit[0], "_", "@"))
+		if err == nil {
+			filePath = strings.Join(filePathSplit[1:], "/")
+		}
+	}
+
+	return filePath
+}
+
+func getPublicKey64() (string, error) {
```
nanjiangshu, line 523: This function looks quite generic, so maybe better to be placed in `helpers.go`.

nanjiangshu (commented):
Looks great! I just added two minor comments regarding the placement of the functions and also potential unit tests.

MalinAhlberg:
Thank you, @nanjiangshu :+1: ! I addressed your suggestions in 3a00170. I did not add any unit test for `AnonymizeFilepath`, though, as I believe it will not be used much in the future, since we are moving away from storing userids as part of the filepath.

nanjiangshu (approved):

download/download.go
MalinAhlberg: Fixed in 3a00170

download/download.go
MalinAhlberg: Fixed in 3a00170

PR #511, opened 2025-02-24
display host_base

### LINKED ISSUE #294: [session] display target site
As a user I would like to see which target my session file was generated from

a/c
- When reading in the session file, (after we have checked the validity of the token #290) we should print out the `host_base` so the user sees which the target site is. This is relevant only when the cli talks with a remote location (download, upload, list etc)
- Display the host base in the message to the user
- Add test(s)

pahatz:
This PR closes #294 .

**Description**
Display `host_base` when running a command that requires a config file (download, list, upload).

**How to test**
Manually, run any command from **download**, **list**, **upload**. They should all display the `host_base`.
Or run the test suite. **list** and **upload** use unit tests while **download** an integration test.

### DIFF
--- a/.github/integration/tests/40_download.sh
+++ b/.github/integration/tests/40_download.sh
@@ -2,7 +2,13 @@
 set -e
 
 # Download file by using the sda-cli download command
-./sda-cli -config testing/s3cmd-download.conf download -dataset-id https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 -outdir test-download main/subfolder/dummy_data.c4gh
+output=$(./sda-cli -config testing/s3cmd-download.conf download -dataset-id https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 -outdir test-download main/subfolder/dummy_data.c4gh)
+
+# Check if we print the host_base in the output
+if [[ $output != *"host_base: localhost:8000"* ]]; then
+    echo "host_base not found in the output"
+    exit 1
+fi
 
 # Check if file exists in the path
 if [ ! -f "test-download/main/subfolder/dummy_data" ]; then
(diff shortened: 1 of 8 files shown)


.github/integration/tests/40_download.sh
```diff
@@ -2,7 +2,13 @@
 set -e
 
 # Download file by using the sda-cli download command
-./sda-cli -config testing/s3cmd-download.conf download -dataset-id https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 -outdir test-download main/subfolder/dummy_data.c4gh
+output=$(./sda-cli -config testing/s3cmd-download.conf download -dataset-id https://doi.example/ty009.sfrrss/600.45asasga -url http://localhost:8080 -outdir test-download main/subfolder/dummy_data.c4gh)
+
+# Check if we print the host_base in the output
+if [[ $output != *"host_base: localhost:8000"* ]]; then
```
nanjiangshu:
```suggestion
if [[ "$output" != *"host_base: localhost:8000"* ]]; then
```

nanjiangshu (commented):
Works well when I tested!
I added a minor comment in string comparison to improve the robustness.

Apart from that, I have a question regarding the word `host_base`. Since this is user facing, do we have a better word then `host_base`. 

I would suggest something like 

```
Remote server (host_base): http://localhost:18000
```

MalinAhlberg (approved):
Very nice!
Addition: I think the suggestion from @nanjiangshu is very good (`Remote server (host_base): ...`).

.github/integration/tests/40_download.sh
pahatz: I opted out of this test after all, but thanks!

pahatz:
I added your suggestion. I noticed some tests were wrong so I fixed those too.

nanjiangshu (approved):
Looks good!

PR #522, opened 2025-03-19
add paging using file marker

### LINKED ISSUE #520: [list] pagination breaks for Ceph backend
**Describe the bug**

Apparently Ceph doesn't support `ListObjectsV2` of aws SDK. This means that the current implementation of list pagination in #502 fails for Ceph S3 backends.

The reason why #502 worked was because it was tested on `minio` S3 that implements V2.

**Steps to reproduce**
Upload 1001 files in a safespring bucket

**Expected behavior**
- [ ] pagination works regardless if S3 backend supports V1 or V2.

**Additional context**

aaperis:
This PR closes #520.

**Description**

The currently used aws-SDK offers two methods for listing buckets: `ListObject` and `ListObjectV2`. The former uses a file `marker`  so that paging can continue from the previously listed S3 object and the later uses a `continuation` token instead. 
`Minio` works with `ListObjectsV2`, meaning that it sends back a `continuationToken` when more than 1000 files are to be listed whereas our current `Ceph` backend sends back file markers instead. Thus, the previously implemented paging using `ListObjectsV2` failed for Ceph.

To make paging more robust, here both methods are added. By default we use `ListObjects`  and if that fails the code repeats the list request using `ListObjectsV2`. Both methods are split out to separate functions so that the code is more readable.

**How to test**
UnitTests remain the same but since the `gofakeS3` library utilizes `ListObjectsV2`, the old tests now also check the new additions, i.e. the code will first try `ListObjects` and if pagination is needed it will revert to using the correct method (see added comment in the testsuite).

Note: all of this will have to change when we migrate to aws-SDK-V2 but at least we'll have (hopefully) a working `sda-cli list` until then.


helpers/helpers.go
```diff
@@ -452,47 +452,35 @@ func ListFiles(config Config, prefix string) ([]*s3.Object, error) {
...
+	params := &s3.ListObjectsInput{
+		Bucket:  aws.String(config.AccessKey + "/"),
+		Prefix:  aws.String(config.AccessKey + "/" + prefix),
+		MaxKeys: aws.Int64(1000),
 	}
-
-	// MaxS3Keys corresponds to the aws SDK MaxKeys that sets the limit of objects
-	// returned with every request. This is only set to >0 for testing purposes,
-	// otherwise the default MaxKeys value (which is 1000) will be used.
+	// This is only set to >0 for unit testing purposes,
```
MalinAhlberg, line 460:
```suggestion
	// This is only set to >0 for unit testing purposes.
```

MalinAhlberg (approved):
Looks good! (although I haven't tested it)

helpers/helpers.go
```diff
@@ -543,3 +531,81 @@ func GetPublicKey64(pubKeyPath *string) (string, error) {
...
+		if result.NextContinuationToken == nil {
+			// Catch the false positive case where the server does not support V2 pagination
+			if *result.IsTruncated {
+				return nil, fmt.Errorf("continuation tokens not supported by backend")
+			}
+
+			break
+		}
+
+		continuationToken = result.NextContinuationToken
```
jbygdell, line 607:
```suggestion
		if result.NextContinuationToken == nil  && *result.IsTruncated {
			// Catch the false positive case where the server does not support V2 pagination
			return nil, fmt.Errorf("continuation tokens not supported by backend")
		}

		fullResult = append(fullResult, result.Contents...)
		continuationToken = result.NextContinuationToken
		if !*result.IsTruncated {
			break
		}
```

helpers/helpers.go
```diff
@@ -543,3 +531,81 @@ func GetPublicKey64(pubKeyPath *string) (string, error) {
...
+		if result.NextMarker == nil {
+			// Catch the false positive case where the server does not support V1 pagination
+			if *result.IsTruncated {
+				return nil, fmt.Errorf("file markers not supported by backend")
+			}
+
+			break
+		}
+
+		marker = result.NextMarker
```
jbygdell, line 568:
```suggestion
		if result.NextMarker == nil && *result.IsTruncated {
			// Catch the false positive case where the server does not support V1 pagination
			return nil, fmt.Errorf("file markers not supported by backend")
		}

		fullResult = append(fullResult, result.Contents...)
		marker = result.NextMarker
		if !*result.IsTruncated {
			break
		}
```

jbygdell (changes requested):

helpers/helpers_test.go
```diff
@@ -562,7 +562,10 @@ func (suite *HelperTests) TestListFiles() {
 	assert.NoError(suite.T(), err, "failed when it shouldn't")
 	assert.Equal(suite.T(), len(result), 1)
 
-	// Test list pagination
+	// Test list pagination.
+	// The used gofakeS3 version utilizes the continuationToken for paging.
+	// Therefore, here we also implicitely test that the ListFiles function
+	// reverts to using ListObjectsV2 when ListObjects paging fails.
```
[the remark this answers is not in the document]
aaperis: I agree. I opened #527 as a placeholder for this, feel free to edit :smile:

jbygdell (approved):

PR #536, opened 2025-04-11
better error message for config flag

### LINKED ISSUE #533: Wrong order of -config produces confusing error
When the config flag is not in the correct position, the error we throw for the user is not clear about what they did wrong.
Currently it gives an error like:

> sda-cli.exe upload -config ./s3cmd.conf -r DATASET_xxxx
> throws this error: flag provided but not defined: -config

Change the error to: "the config flag should come before the subcommand (upload)" and print an example

MalinAhlberg:
This PR closes #533 

**Description**
The current output looks like this
```
$ ./sda-cli  upload katt.txt.c4gh -config /tmp/shared/s3cfg 
Error: failed parsing arguments, reason: the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg upload'
```

If you wonder about the implementation:
- 89873c4d2ba97b74e22ea6d8874327320c0fdbe7  fixes the problematic behavior. Unfortunately, in a not supernice way (code wise). The `flag` package insists on printing it's own error (`flag provided but not defined: -config`) and then all the help messages about the current subcommand, before you can make it print the error you want:
    ```
    ./sda-cli  upload ../sensitive-data-anchive/katt.txt.c4gh -config /tmp/shared/s3cfg 
    flag provided but not defined: -config
    Usage of upload:
      -accessToken ACCESSTOKEN
    
    ...
    
    Error: failed parsing arguments, reason: the config flag should come before the 
    subcommand. Eg 'sda-cli -config s3cfg upload'
    ```
   Therefore, the current solution checks the args before they are parsed.

- 2652a5d413921099bd832633e0343fdd1dd725c1 is not actually doing anything for this solution, but `ContinueOnError` is the correct flag to use if you want the [calling function of `Parse`](https://github.com/NBISweden/sda-cli/blob/5d978f687304bb640999bf02d643489e73934f14/helpers/helpers.go#L172) to be able to see the error produced.

- 5d978f687304bb640999bf02d643489e73934f14 adds tests.

**How to test**
Try the different subcommands, with the config flag given after the subcommand:
```
$ ./sda-cli  upload katt.txt.c4gh -config /tmp/shared/s3cfg 
Error: failed parsing arguments, reason: the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg upload'
```

```
$./sda-cli list --datasets -url http://localhost:8080 -config  testing/s3cmd.conf
Error: failed parsing arguments, reason: the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg list'
```

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -164,6 +164,11 @@ func ParseArgs(args []string, argFlags *flag.FlagSet) error {
 	pos, args = getPositional(args)
 	// append positional args back at the end of args
 	args = append(args, pos...)
+	for _, item := range args {
+		if item == "-config" || item == "--config" {
+			return fmt.Errorf("the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg %s'", argFlags.Name())
+		}
+	}
 	err := argFlags.Parse(args[1:])
 
 	return err
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -60,7 +60,7 @@ Arguments:
 
 // Args is a flagset that needs to be exported so that it can be written to the
 // main program help
-var Args = flag.NewFlagSet("upload", flag.ExitOnError)
+var Args = flag.NewFlagSet("upload", flag.ContinueOnError)
 
 var forceUnencrypted = Args.Bool("force-unencrypted", false, "Force uploading unencrypted files.")
 
--- a/.github/integration/tests/10_encrypt_decrypt.sh
+++ b/.github/integration/tests/10_encrypt_decrypt.sh
@@ -104,4 +104,16 @@ done
 
 echo "Could decrypt with all keys from concatenated key"
 
+if ./sda-cli decrypt -key $key.sec.pem data_file_keys.c4gh -config  testing/s3cmd.conf 2> >(grep "the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg decrypt" > /dev/null ) -ne 1
+then
+    echo "Unexpected error message"
+    exit 1
+fi
+if ./sda-cli encrypt -key $key.sec.pem data_file_keys.c4gh -config  testing/s3cmd.conf 2> >(grep "the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg encrypt" > /dev/null ) -ne 1
+then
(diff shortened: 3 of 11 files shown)


upload/upload.go
```diff
@@ -60,7 +60,7 @@ Arguments:
 
 // Args is a flagset that needs to be exported so that it can be written to the
 // main program help
-var Args = flag.NewFlagSet("upload", flag.ExitOnError)
```
pahatz, line 63: Shouldn't we do the same for `list` and `download`?

pahatz (commented):
Nice fix, great description! I have a comment for the fix for the rest of the commands that use a config flag.

upload/upload.go
MalinAhlberg: yes, good catch. Updated in f218e32, so that all callers of `helpers.ParseArgs` uses `ContinueOnError`.

pahatz (approved):

helpers/helpers.go
```diff
@@ -164,6 +164,11 @@ func ParseArgs(args []string, argFlags *flag.FlagSet) error {
 	pos, args = getPositional(args)
 	// append positional args back at the end of args
 	args = append(args, pos...)
+	for _, item := range args {
+		if item == "-config" || item == "--config" {
+			return fmt.Errorf("the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg %s'", argFlags.Name())
+		}
+	}
```
jbygdell, line 171: There are multiple ways of doing this below is another example that uses the slice index.
Both methods are equally valid but here we don't need to iterate through the entire slice of arguments.
```
	if idx := slices.IndexFunc(args, func(arg string) bool { return strings.Contains(arg, "-config") }); idx != 0 {
		return fmt.Errorf("the config flag should come before the subcommand. Eg 'sda-cli -config s3cfg %s'", argFlags.Name())
	}
```

jbygdell (approved):

helpers/helpers.go
MalinAhlberg: Thanks, that's good to know!

I will go with the current version, since it is approved, tested etc. But very nice to know for the future.

PR #542, opened 2025-04-24
Feature/download resume dataset

### LINKED ISSUE #405: [download] Resume download of a dataset
As a user, I want to be able to resume downloading a dataset so that I don't have to download all the files again if my download job is interrupted.

Ad a `--continue` flag to do this (or `--resume` would be fine as well)

**Acceptance criteria**
- [x] Download of files that exist locally is skipped provided that these have the same name.
- [x] Tests verifying the changes are added

**Additional context**

aaperis:
This PR closes #405.

**Description**

If the flag `-continue` is given to `download` then any existing file with the same name and file path will be skipped during download.

**How to test**
See that the added integration tests pass.

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -45,6 +45,7 @@ Required options:
   -url <uri>                  The url of the download server.
 
 Optional options:
+  -continue                   Skip already downloaded files and continue with downloading the rest.
   -pubkey <public-key-file>   Key to use for encrypting downloaded files server-side.
                               This key must be given here or in the config file.
   -outdir <dir>               Directory to save the downloaded files.
@@ -77,6 +78,8 @@ var recursiveDownload = Args.Bool("recursive", false, "Download content of the f
 var fromFile = Args.Bool("from-file", false, "Download files from file list.")
 
 var pubKeyBase64 string
+
+var continueDownload = Args.Bool("continue", false, "Skip existing files and continue with the rest.")
 
 // necessary for mocking in testing
 var getResponseBody = getBody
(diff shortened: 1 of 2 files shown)


download/download.go
```diff
@@ -327,6 +330,15 @@ func downloadFile(uri, token, pubKeyBase64, filePath string) error {
...
 		filePath += ".c4gh"
 	}
+
+	if *continueDownload {
+		if _, err := os.Stat(filePath); !errors.Is(err, os.ErrNotExist) {
+			fmt.Printf("Skipping download to %s, file already exists\n", filePath)
+
+			return nil
+		}
+	}
```
jbygdell, line 340:
```suggestion
	if _, err := os.Stat(filePath); !errors.Is(err, os.ErrNotExist) && *continueDownload {
		fmt.Printf("Skipping download of %s, file already exists\n", filePath)

		return nil
	}
```

It should be doable without the nested ifs.
Also should the message be to stdout or stderr?

aaperis: The nested if was a conscious choice because this way we spare the unnecessary calls to `os.Stat` for every file if `-continue` is false (imagine 50000 files for example). 

As for the second part, I see that we use a mixture of stderr and stdout for our messages. For example, `Upload` outputs files that are skipped to stdout whereas `decrypt` to stderr, the later has a `Warning:` prefix as well. So I am inclined to leave it as is and open an issue to harmonize all relevant messages.

jbygdell: Fair enough.

MalinAhlberg (approved):
All looks well to me, and works fine when tested.

I get a bit confused by the name `continue`, however. It sounds like that without this flag, the program would *stop* when trying to download files that already exists. Maybe we could make this clearer in the docs? (sorry for not realizing and raising this earlier)

aaperis:
> All looks well to me, and works fine when tested.
> 
> I get a bit confused by the name `continue`, however. It sounds like that without this flag, the program would _stop_ when trying to download files that already exists. Maybe we could make this clearer in the docs? (sorry for not realizing and raising this earlier)

I can see why it may seem confusing and I agree that this is not ideal. I wanted to avoid using `-resume` because this could be perceived as "resuming a partially downloaded file" whereas here this is not the case. We also use `-continue` for a similar functionality in the `Upload` module so I thought this would be at least consistent.
Given that this is the first iteration for implementing this functionality (I would prefer to check more than the filenames), I say we go with it for now, and I open an issue for a follow-up that includes the naming of the flag (#544).

Issue #460, opened 2024-10-24
[sda-cli] Provide feedback more promptly when recursively uploading a folder with many large files


nanjiangshu:
**Please describe the feature**
As a user, I want sda-cli to provide feedback more promptly when recursively uploading a folder with many large files, so that I can understand the program is running properly.

Currently, the `upload` sub-command encrypts all files before uploading, and users may encounter situations where the program seems to hang for many hours.

Print a message when starting "Encrypting all files before uploading".

**Acceptance criteria**
- [ ] A message is printed in the terminal that all files are encrypted before uploading.

**Additional context**

**Estimation of size**: small

**Estimation of priority**: medium

PR #552, opened 2025-06-02
[encrypt] adding progress bar to file encryption

zeidlitz:
This PR closes [#460 ].

**Description**
In order to give clearer feedback to the user when running sda-cli there was a need to understand the progress of file encryption. This should (hopefully) align with how progress is tracked in for example upload. 

**How to test**
sda-cli upload

### DIFF
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -21,6 +21,8 @@ import (
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/neicnordic/crypt4gh/streaming"
 	log "github.com/sirupsen/logrus"
+	"github.com/vbauerster/mpb/v8"
+	"github.com/vbauerster/mpb/v8/decor"
 )
 
 // Help text and command line flags.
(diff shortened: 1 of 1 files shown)


nanjiangshu (commented):
Well done @zeidlitz!
The original description of the issue is just to show a message like `Encrypting files before uploading...`. The reason to do that is because when the file size is large, e.g. 20GB, it will take a long time before it shows any progress information to the user. 
Now you added a progress bar to show the progress promptly. I think this is also a nice solution. 

However, there are few minor issues:

1. It will be good to show the name of the file that is encrypting, similar as for the uploading progress bar. For example, show

File <filename>: encrypting ...

2. It seems the size of the file is not showing correctly

3. Since this PR implements a few feature, it is recommend to use a more informative branch name, e.g. `feature/balbal` instead of a general `develop` branch. By the way, we don't have a `develop` branch for this repo.

zeidlitz:
Thanks for the input @nanjiangshu 👍 I will make adjustments and publish a new suggestion

encrypt/encrypt.go
```diff
@@ -225,16 +227,34 @@ func Encrypt(args []string) error {
...
+		fileSize, _ := helpers.GetFileSize(file.Unencrypted)
+		bar := p.AddBar(fileSize,
+			mpb.PrependDecorators(
+				decor.Name(fmt.Sprintf("Encrypting %-12s", file.Unencrypted), decor.WCSyncWidth),
+			),
+			mpb.AppendDecorators(
+				decor.CountersKibiByte("% .1f / % .1f"),
+				decor.OnComplete(decor.Percentage(decor.WC{W: 5}), "done"),
+			),
+		)
```
jbygdell, line 247: If you want to track the encryption progress this should be in the unexported function starting down on line 422.
You need to implement a progress tracker in combination with a teereader or multiwriter to achieve the goal.

jbygdell (approved):

nanjiangshu (approved):
Looks good.

PR #560, opened 2025-08-15
Feature/migrate to aws sdk v2 (#479)

### LINKED ISSUE #479: Migrate from github.com/aws/aws-sdk-go to github.com/aws/aws-sdk-go-v2
As a developer, I want the code to use the latest version of a module so that we can make use of the new functionality and security options.

This has already been done in the sensitive-data-archive repository, so look there for inspiration.

**Acceptance criteria**
- [ ] everything works as before

**Additional context**
See [here](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/) for why.

KarlG-nbis:

This PR closes #479 

**Description**
Migrates from `aws-sdk-go` to `aws-sdk-go-v2`

**How to test**
Run unit test to verify existing behaviour is not affected


helpers/helpers.go
```diff
@@ -396,34 +398,44 @@ func CheckTokenExpiration(accessToken string) error {
...
-		Region: aws.String("us-west-2"),
-		Credentials: credentials.NewStaticCredentials(
+func ListFiles(config Config, prefix string) ([]types.Object, error) {
+	s3Endpoint := config.HostBase
+	if !strings.HasPrefix(s3Endpoint, "http") {
+		s3Endpoint = fmt.Sprintf("http://%s", config.HostBase)
+		if !config.UseHTTPS {
+			s3Endpoint = fmt.Sprintf("https://%s", config.HostBase)
+		}
+	}
```
jbygdell, line 408:
```suggestion
```
This is handled in the `LoadConfig` function.
Although it is not perfect since it makes a bunch of assumptions on how the entry will look like.


If you use the "net/url" package to handle the hostbase on line 243 it can look something like this.
```
	u, err := url.Parse(config.HostBase)
	if err != nil {
		return nil, fmt.Errorf("hostbase parsing error: %v", err)
	}

	switch {
	case u.Scheme == "https":
		config.UseHTTPS = true
	case config.UseHTTPS:
		u.Scheme = "https"
	default:
		u.Scheme = "http"
		config.UseHTTPS = false
	}
```
The tests need to use `ts.URL` as the url since IP addresses needs the scheme in order to parse properly.
Then all these cases will be viable:
- http://127.0.0.1:32564
- http://example.com
- https://example.com
- example.com:8080
- example.com

upload/upload.go
```diff
@@ -126,23 +130,34 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
 		}
 		_ = f.Close()
 	}
+	s3Endpoint := config.HostBase
+	if !strings.HasPrefix(s3Endpoint, "http") {
+		s3Endpoint = fmt.Sprintf("http://%s", config.HostBase)
+		if !config.UseHTTPS {
+			s3Endpoint = fmt.Sprintf("https://%s", config.HostBase)
+		}
+	}
```
jbygdell, line 139:
```suggestion
```

This is handled in the LoadConfig function.
Although it is not perfect since it makes a bunch of assumptions on how the entry will look like.

jbygdell (changes requested):

helpers/helpers.go
```diff
@@ -396,34 +398,44 @@ func CheckTokenExpiration(accessToken string) error {
...
-func ListFiles(config Config, prefix string) ([]*s3.Object, error) {
-	sess := session.Must(session.NewSession(&aws.Config{
-		// The region for the backend is always the specified one
-		// and not present in the configuration from auth - hardcoded
-		Region: aws.String("us-west-2"),
-		Credentials: credentials.NewStaticCredentials(
+func ListFiles(config Config, prefix string) ([]types.Object, error) {
+	s3Endpoint := config.HostBase
+	if !strings.HasPrefix(s3Endpoint, "http") {
+		s3Endpoint = fmt.Sprintf("http://%s", config.HostBase)
```
nanjiangshu, line 404:
```suggestion
		if config.UseHTTPS {
			s3Endpoint = fmt.Sprintf("https://%s", config.HostBase)
		} else {
			s3Endpoint = fmt.Sprintf("http://%s", config.HostBase)
		}   
```
Current code overwrite the prefix to `https://`  only when `UseHTTPS` is `false`

upload/upload.go
nanjiangshu, line 139: This has been done at https://github.com/NBISweden/sda-cli/blob/main/helpers/helpers.go#L240-L242. If you want to improve the logic, modify there.

nanjiangshu (commented):
Looks good in general and all major functionalities seems to work as before. 
Just a comment for how `HostBase` is configured.

helpers/helpers.go
```diff
@@ -237,8 +240,30 @@ func LoadConfigFile(path string) (*Config, error) {
...
+	config.HostBase = u.String()
+
+	// Parse URL again to validate that a host can be parsed after scheme has been enforced
+	u, err = url.Parse(config.HostBase)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse host base from configuration file, reason: %v", err)
+	}
+	if u.Host == "" {
+		return nil, fmt.Errorf("failed to parse host base from configuration file, reason: a valid host can not be parsed")
 	}
```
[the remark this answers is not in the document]
KarlG-nbis: Yeah the reason I put it afterwards with another `url.Parse` is that when input was "example.com" there is no host, and the `u.Path` is "example.com", but after enforcing the scheme, it parses "example.com" as the host. The other alternative is not allowing formats without scheme such as "example.com"

image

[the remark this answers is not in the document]
jbygdell: That's interesting.
maybe if we do like this then:

```
case u.Host == "" && u.Path == "":
	return error
case u.Host == "" && u.Path != "":
	u.Host = u.Path
	u.Path = ""
```

PR #566, opened 2025-08-20
Feature/decrypt unit test refactor (#466)

### LINKED ISSUE #466: Refactor decryption tests
As a developer, I want the tests to be limited in scope so that it is easy to see where a failure occurs, that includes having proper names for the tests.

**Acceptance criteria**
- [ ] Ensure all tests can be run in isolation
- [ ] Ensure all tests has relevance.
- [ ] Ensure all temp files are removed after the test has run.
- [ ] Split tests with multiple checks into separate tests.
- [ ] Each test file should only contain one `test suite`.
- [ ] Thing that can be shared between tests should be set up during the `SetupSuite` step.

**Additional context**

**Estimation of size**: small

KarlG-nbis:
# Related issue(s) and PR(s)
This PR closes #466

# Description
Refactors the decrypt unit test to
- [X] Ensure all tests can be run in isolation
- [X] Ensure all tests has relevance.
- [X] Ensure all temp files are removed after the test has run.
- [X] Split tests with multiple checks into separate tests.
- [X] Each test file should only contain one test suite.
- [X] Thing that can be shared between tests should be set up during the SetupSuite step.

# How to test
Run unit tests

### DIFF
--- a/decrypt/decrypt_test.go
+++ b/decrypt/decrypt_test.go
@@ -68,7 +68,7 @@ func (suite *DecryptTestSuite) createNewEncryptedFile() {
 	if err != nil {
 		suite.FailNow("failed to create test file in temporary directory", err)
 	}
-	fileContent := []byte(fmt.Sprintf("This is some fine content right here, in file: %s", testFile.Name()))
+	fileContent := fmt.Appendf([]byte{}, "This is some fine content right here, in file: %s", testFile.Name())
 	// ... and write the known content to it
 	err = os.WriteFile(testFile.Name(), fileContent, 0600)
 	if err != nil {
@@ -254,7 +254,7 @@ func (suite *DecryptTestSuite) TestDecryptWithCleanArgSuccess() {
 }
 
 func (suite *DecryptTestSuite) TestDecryptWithCleanArgWrongPassword() {
-	os.Setenv("C4GH_PASSWORD", "wrong")
+	_ = os.Setenv("C4GH_PASSWORD", "wrong")
 
 	err := Decrypt([]string{
 		"decrypt",
@@ -294,7 +294,7 @@ func (suite *DecryptTestSuite) TestDecryptMultipleFilesSuccess() {
 	assert.NoError(suite.T(), err)
 
 	// Check the decrypted files
-	for _, file := range suite.testFiles { // Check file2.txt and file3.txt
+	for _, file := range suite.testFiles {
 
 		// Check that the encrypted file was remains
 		_, err = os.Stat(file.encryptedFileName)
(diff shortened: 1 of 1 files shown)


zeidlitz (approved):

encrypt/encrypt.go
```diff
@@ -78,6 +78,12 @@ func init() {
 	})
 }
 
+// EmptyPublicKeyFileList cleans the publicKeyFileList
+// main use case is in unit tests when files are to be encrypted with different keys in different invocations
+func EmptyPublicKeyFileList() {
+	publicKeyFileList = make([]string, 0)
+}
+
```
jbygdell, line 86: A bit odd location for this one. It could easily be placed in the `decrypt_test` file and still be accessible where it is used.

jbygdell (changes requested):
looks good. A negative test for decrypting with the wrong private key should probably be added.

encrypt/encrypt.go
KarlG-nbis: Well the issue was that this is in the `encrypt` package that is used to encrypt the dummy files created during the decrypt tests, and `publicKeyFileList` is a private variable, so needs to belong in the `encrypt` package, but could move it to the `encrypt_test`, bottom of the file, new file, etc

KarlG-nbis: Moved the func definition to lower in the `encrypt.go` file

aaperis:
Is there a functional reason of splitting the testsuite to multiple small files instead of having everything in the `decrypt_test.go` ?

jbygdell (commented):
In order to not have to duplicate to much code try to follow the following guides:
* all test files for a package belong to the same suite.
* use `setupSuite` to generate stuff that is used by all tests
* use `teardownSuite` to remove shared files at the end.
* use `setupTest` for things that needs to be reset between each test
* use `teardownTest` to remove test files/folders created by a test step

KarlG-nbis:
> In order to not have to duplicate to much code try to follow the following guides:
> 
> * all test files for a package belong to the same suite.
> * use `setupSuite` to generate stuff that is used by all tests
> * use `teardownSuite` to remove shared files at the end.
> * use `setupTest` for things that needs to be reset between each test
> * use `teardownTest` to remove test files/folders created by a test step

Main reason for the duplication of code, etc was that I wanted to introduce seperation of responsibility, eg that the EncryptTestSuite prepares what is needed and tests the `Decrypt` function, TestReadPrivateKeyFile prepares what is needed and tests the ReadPrivateKeyFile function. In my opinion this seperation of responsibility makes the tests more readable and maintainable given there is a clearer definition of what a test suite / test function aims to tests

jbygdell:
> Main reason for the duplication of code, etc was that I wanted to introduce seperation of responsibility, eg that the EncryptTestSuite prepares what is needed and tests the `Decrypt` function, TestReadPrivateKeyFile prepares what is needed and tests the ReadPrivateKeyFile function. In my opinion this seperation of responsibility makes the tests more readable and maintainable given there is a clearer definition of what a test suite / test function aims to tests

Based on how the code look I can understand why you have done it. But if you look a bit closer you will see that it is possible to simplify the code to the extent that all this separation will not be needed.

PR #568, opened 2025-08-22
Feature/download unit test refactor

### LINKED ISSUE #467: Refactor download tests
As a developer, I want the tests to be limited in scope so that it is easy to see where a failure occurs, that includes having proper names for the tests.

**Acceptance criteria**
- [ ] Ensure all tests can be run in isolation
- [ ] Ensure all tests has relevance.
- [ ] Ensure all temp files are removed after the test has run.
- [ ] Split tests with multiple checks into separate tests.
- [ ] Each test file should only contain one `test suite`.
- [ ] Thing that can be shared between tests should be set up during the `SetupSuite` step.

**Additional context**

**Estimation of size**: small

KarlG-nbis:
# Related issue(s) and PR(s)
This PR closes #467 

# Description
Refactors the download unit test to
- [X] Ensure all tests can be run in isolation
- [X] Ensure all tests has relevance.
- [X] Ensure all temp files are removed after the test has run.
- [X] Split tests with multiple checks into separate tests.
- [X] Each test file should only contain one test suite.
- [X] Thing that can be shared between tests should be set up during the SetupSuite step.

# How to test
Run unit tests

### DIFF
--- a/download/download_test.go
+++ b/download/download_test.go
@@ -64,23 +64,16 @@ func (suite *DownloadTestSuite) SetupSuite() {
         	]`)
 
 		case "/s3/TES01/files/dummy-file.txt.c4gh":
-
 			// Set the response status code
 			w.WriteHeader(http.StatusOK)
-
 			fmt.Fprint(w, "test content dummy file")
-
 		case "/s3/TES01/files/file1.c4gh":
-
 			// Set the response status code
 			w.WriteHeader(http.StatusOK)
-
 			fmt.Fprint(w, "test content file 1")
 		case "/s3/TES01/files/file2.c4gh":
-
 			// Set the response status code
 			w.WriteHeader(http.StatusOK)
-
 			fmt.Fprint(w, "test content file 2")
 		case "/metadata/datasets":
 			// Set the response status code


aaperis:
> > Is there a functional reason of splitting the testsuite to multiple small files?
> 
> Main reason is just for readability, given the different files are running tests on different functions. Main functional reason would be that some of what we setup for each test in the DownloadTestSuite is not needed for the TestGetDatasets, etc but again that could still be in the same file, so no real functional reasoning for splitting the files.

Then since there is no functional need, I see no reason to break the pattern `module.go`, `module_test.go` that we follow throughout our repos. 
We could of course open a discussion of how we would prefer to structure our tests where we can all have a say :-) .

aaperis:
> > > Looks great! I like these tests to be spitted in separate files.
> > 
> > 
> > Letting aside my personal taste regarding this, before merging we should agree to adopt this file structure for **all the repos** we work with. So that there is some consistency at least in the future.
> 
> In that case we also need a proper naming scheme for the files, like `packageName_functionalityToTest_test.go`. Also the `packageName_test.go` file should contain the `suite setupSuite` and `setupTest` as well as all shared components for the tests so we don't get the same stuff duplicated in the separate test files.

Yes, this makes sense. We could also add these in our contribution guidelines. Let's briefly discuss it at standup.

download/download_test.go
```diff
@@ -1,324 +1,483 @@
...
+			w.WriteHeader(http.StatusOK)
+
+			fmt.Fprint(w, "test content file 1")
+		case "/s3/TES01/files/file2.c4gh":
+
+			// Set the response status code
+			w.WriteHeader(http.StatusOK)
+
+			fmt.Fprint(w, "test content file 2")
+		case "/metadata/datasets":
```
jbygdell, line 85:
```suggestion
		case "/s3/TES01/files/dummy-file.txt.c4gh":
			// Set the response status code
			w.WriteHeader(http.StatusOK)
			fmt.Fprint(w, "test content dummy file")
		case "/s3/TES01/files/file1.c4gh":
			// Set the response status code
			w.WriteHeader(http.StatusOK)
			fmt.Fprint(w, "test content file 1")
		case "/s3/TES01/files/file2.c4gh":
			// Set the response status code
			w.WriteHeader(http.StatusOK)
			fmt.Fprint(w, "test content file 2")
		case "/metadata/datasets":
```

Just some nitpicking, but try to keep whitespace usage consistent.

PR #588, opened 2025-09-01
Add linter config file

jbygdell:
**Description**
Copy the config  file from the SDA repo so we get an unified handling of the linter between the repositories.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,63 @@
+version: "2"
+formatters:
+  enable:
+    - gofmt
+    - goimports
+linters:
+  default: none
+  enable:
+    - bodyclose
+    - gocritic
+    - gosec
+    - govet
+    - nestif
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -0,0 +1,63 @@
...
+    - nestif
+    - nlreturn
+    - revive
+    - rowserrcheck
+    - staticcheck
+  settings:
+    gosec:
+      excludes:
+        - G104
+        - G115
```
nanjiangshu, line 22: Why make this difference for `G115` compared to the SDA repo?

jbygdell: G115 should be in the default exclude list because max(int32) < max(uint64).
This should be added to the SDA repo as well.

zeidlitz (approved):

kostas-kou (approved):

Issue #563, opened 2025-08-18
[sda-cli] Migrate to a cli parsing library


nanjiangshu:
# Describe the bug
The `-continue` flag of the `upload` subcommand does not work at all positions.

```sh
# These works
../sda-cli -config s3cmd-bp-staging.conf upload  -encrypt-with-key mykey.pub.pem  -r test_dir1  -continue
../sda-cli -config s3cmd-bp-staging.conf upload  -encrypt-with-key mykey.pub.pem -continue  -r test_dir1   -targetDir test_dir1_try5 

# These are not working
../sda-cli -config s3cmd-bp-staging.conf upload -continue -encrypt-with-key mykey.pub.pem  -r test_dir1
../sda-cli -config s3cmd-bp-staging.conf upload -encrypt-with-key mykey.pub.pem  -r test_dir1 -continue  -targetDir test_dir1_try5
```

# Proposed solution
Migrate command line handling to a library. Cobra is one candidate.

# Steps to reproduce

# Expected behavior
- [ ] Tests verifying the fix are added
- [ ] The `-continue` flag should work at all positions as long it meets the requirement of the command line syntax

# Additional context
See also https://github.com/NBISweden/sda-cli/issues/557

**Estimation of size**: medium

**Estimation of priority**: medium

PR #602, opened 2025-09-24
[refactoring] Add spf13/cobra library to handle argument parsing

### LINKED ISSUE #430: [list] Incosistent behavior when calling sda-cli list with no arguments
Check this after #563 has been merged.

```
./sda-cli list
Error: failed to load config file, reason: failed to read the configuration file
```
instead of `Usage info` like the rest of the sda-cli commands.

zeidlitz:
This change wants to address #430 and #563 and generally improve how handling of input arguments is done.

**Description**
There is a lost of logic in place dealing with arguments, flags, commands, sub-commands and their respective flags. There is also some inconsistencies and flawed logic in the current logic. By adding this library we can address a lot of these issues and hopefully have easier maintainability.

This PR closes #430 & #563


cmd/root.go
```diff
@@ -0,0 +1,44 @@
...
+
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
+	rootCmd.PersistentFlags().StringP("config", "c", "s3cmd.conf", "The configuration file for s3cmd")
```
KarlG-nbis, line 43: we could also mark it as required with `rootCmd.MarkPersistentFlagRequired("config")` since it should be required for all commands

main.go
```diff
@@ -1,211 +1,7 @@
...
-
-		// print main help
-		fmt.Println(Usage)
-
-		// Print link to the README
-		readmeURL := "https://github.com/NBISweden/sda-cli/blob/main/README.md"
-		if !strings.Contains(Version, "development") {
-			readmeURL = fmt.Sprintf("https://github.com/NBISweden/sda-cli/blob/%s/README.md", Version)
-		}
-		fmt.Println("For more information, see the README at:", readmeURL)
```
KarlG-nbis, line 198: This information is no longer part of the help message

cmd/list.go
```diff
@@ -0,0 +1,40 @@
...
+func init() {
+	rootCmd.AddCommand(listCmd)
+	BytesFormat = listCmd.Flags().Bool("bytes", false, "Display file sizes in bytes instead of human-readable format")
+	Dataset = listCmd.Flags().String("dataset", "", "List all files in the sepcified dataset, including dataset size")
+	Datasets = listCmd.Flags().String("datasets", "", "List all datasets available in the user's folder")
+	URL = listCmd.Flags().String("url", "", "Specify the SDA download server URL")
+	Prefix = listCmd.Flags().String("prefix", "", "Optional prefix to filter results to a specific location or folder path")
+}
+
+var listCmd = &cobra.Command{
```
KarlG-nbis, line 23: imo it would be cleaner to have this listCmd in the list package such that all that relates to the list cmd is in that package, but not sure what the cobra best practice is

cmd/list.go
```diff
@@ -0,0 +1,40 @@
+package cmd
+
+import (
+	"github.com/NBISweden/sda-cli/list"
+	"github.com/spf13/cobra"
+)
+
+var Dataset *string
```
KarlG-nbis, line 8: given these flag values have default values and therefor would never be nil, to me it would make more sense to have it as non pointers type, eg by

```
var dataset string

func init() {
  listCmd.Flags().StringVar(&dataset, "dataset", "", "List all files in the sepcified dataset, including dataset size")
}
``` 

Also do these need to be exported/public?

cmd/root.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/spf13/cobra"
+)
+
+var rootCmd = &cobra.Command{
+	Use:   "sda-cli",
+	Short: "CLI tool used to interact with SDA",
+	Long: `Command line tool used to interact with the Sensitive Data Archive (SDA)
+
+Commands:
+  createKey   Creates a Crypt4GH key pair
```
KarlG-nbis, line 15: In the long message we shouldnt need to manually write the commands and flags supported, that should be handled by cobra

cmd/list.go
```diff
@@ -0,0 +1,40 @@
...
+var Prefix *string
+var BytesFormat *bool
+
+func init() {
+	rootCmd.AddCommand(listCmd)
+	BytesFormat = listCmd.Flags().Bool("bytes", false, "Display file sizes in bytes instead of human-readable format")
+	Dataset = listCmd.Flags().String("dataset", "", "List all files in the sepcified dataset, including dataset size")
+	Datasets = listCmd.Flags().String("datasets", "", "List all datasets available in the user's folder")
+	URL = listCmd.Flags().String("url", "", "Specify the SDA download server URL")
+	Prefix = listCmd.Flags().String("prefix", "", "Optional prefix to filter results to a specific location or folder path")
```
KarlG-nbis, line 20: Now we we changing the behaviour of the list command with expecting the prefix as a flag instead of an arg

nanjiangshu (approved):
Looks good. I don't have more comments. Great job.
As to the version, we can tag it with 0.4.x after merging the PR https://github.com/NBISweden/sda-cli/pull/610 (which will be something like 0.3.x)

version/version.go
```diff
@@ -2,53 +2,55 @@ package version
...
 	"os"
 	"time"
 
+	"github.com/NBISweden/sda-cli/cmd"
 	"github.com/hashicorp/go-version"
+	"github.com/spf13/cobra"
 )
 
-// Help text and command line flags.
+var Version = "0-development"
```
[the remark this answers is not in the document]
zeidlitz: Done!

PR #619, opened 2025-11-04
Fix display of upload file location

viklund:
The call to shutdown the progress bar (p.Shutdown) clears the last line of output, causing the text "file uploaded to %s" to be removed and instead the finished progressbar is shown twice.

Moving the Printf to after p.Shutdown fixes this.

This PR closes [issue number].

**Description**

**How to test**


zeidlitz (approved):

nanjiangshu (approved):
Nice fix!

PR #632, opened 2025-12-10
feature/ Add encryptionScheme=C4GH when calling htsget-rs

### LINKED ISSUE #631: [htsget] Add encryptionScheme=C4GH when calling htsget-rs
# Description
htsget-rs has been updated to support c4gh encrypted index and gzi files, for the htsget-rs to expect the files to be encrypted in such a format it expects the encryptionScheme=C4GH parameter to be present in the http request

# A/Cs
- when sda-cli htsget cmd calls htsget-rs encryptionScheme=C4GH parameter is part of the request

KarlG-nbis:
This PR closes #631 .

**Description**
Adds encryptionScheme=C4GH when calling htsget-rs

### DIFF
--- a/htsget/htsget.go
+++ b/htsget/htsget.go
@@ -73,7 +73,7 @@ type HtsgetResponse struct {
 	Htsget HtsgetInfo `json:"htsget"`
 }
 
-func Htsget(args []string, configPath string) error {
+func Htsget(_ []string, configPath string) error {
 	if datasetID == "" || fileName == "" || htsgetHost == "" || publicKeyFile == "" {
 		return errors.New("missing required arguments, dataset, filename, host and key are required")
 	}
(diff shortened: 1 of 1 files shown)


htsget/htsget.go
```diff
@@ -91,9 +91,9 @@ func Htsget(args []string, configPath string) error {
 
 	// TODO: Add cases for different type of files
 	// i.e. bam files require the /reads/, replace for vcf
-	url := htsgetHost + "/reads/" + datasetID + "/" + fileName
+	url := htsgetHost + "/reads/" + datasetID + "/" + fileName + "?encryptionScheme=C4GH"
 	if referenceName != "" {
-		url = url + "?referenceName=" + referenceName
+		url = url + "&referenceName=" + referenceName
```
zeidlitz: what's the reason for: "?referenceName=" -> "&referenceName=" ?

just curious

zeidlitz (approved):
LGTM

htsget/htsget.go
KarlG-nbis: as before this was the only parameter, so if it was appended we needed the `?` to give that indication, now the `?` is always appended when adding the `encryptionScheme=C4GH` on line 94, so if we add `referenceName` parameter as well, we just use `&` to add it as another paramater

tl:dr, `?` indicates when the parameters start in a URL, and `&` is the delimiter between parameters

zeidlitz (approved):

PR #637, opened 2026-01-15
download returns useful message when HTTP status code is 412

### LINKED ISSUE #623: should show actual error message instead of just 412 when cli version does not meet the requirement
# Describe the bug
When the CLI version does not match the requirement. Users get the error message

`Error: failed to get datasets, reason: server returned status 412`

This error message is not very informative.

It should show the original error message. e..g 

`Error: Your sda-cli client version is outdated, please update to at least version 'v0.3.0'.`

# Steps to reproduce
Run the following command
```bash
./sda-cli-v0.2.1 --config s3cmd-bp-download.conf list  --url https://download.bp.nbis.se  --datasets 
```
You get
```
Error: failed to get datasets, reason: server returned status 412
```
# Expected behavior
- [ ] SDA-CLI show the informative error message when the CLI version does not meet the requirement

# Demo

Run an old version of sda-cli, show the incorrect message. Run the new version, show the correct message.

nanjiangshu:
This PR closes #623 

**Description**
sda-cli returns the original message when the HTTP status code is 412. When the status code is others, still return just the status code to avoid exposing unnecessary information to the user

**How to test**
That unit test and integration tests pass.

### DIFF
--- a/.github/integration/tests/40_download.sh
+++ b/.github/integration/tests/40_download.sh
@@ -166,13 +166,12 @@ error_output=$(./sda-cli-v0.1.0 --config testing/s3cmd-download.conf download \
     --url http://localhost:8080 \
     --outdir test-version-fail main/subfolder/dummy_data.c4gh 2>&1 || true)
 
-echo "Captured Output: $error_output"
-
 # Check if the error message contains the specific string returned by the server (412 logic)
 expected_msg="Your sda-cli client version is outdated, please update to at least version"
 
 if [[ "$error_output" != *"$expected_msg"* ]]; then
     echo "sda-cli does not show correct version error message."
+    echo "Captured Output: $error_output"
     echo "Expected string: $expected_msg"
     exit 1
 fi


.github/integration/tests/40_download.sh
```diff
@@ -154,8 +154,31 @@ else
...
+
+# We expect this to fail because the server requires v0.2.0 or higher
+error_output=$(./sda-cli-v0.1.0 --config testing/s3cmd-download.conf download \
+    --pubkey user_key.pub.pem \
+    --dataset-id https://doi.example/ty009.sfrrss/600.45asasga \
+    --url http://localhost:8080 \
+    --outdir test-version-fail main/subfolder/dummy_data.c4gh 2>&1 || true)
+
+echo "Captured Output: $error_output"
+
```
jbygdell:
```suggestion
```

download/download.go
```diff
@@ -421,19 +423,21 @@ func getBody(requestURL, token, pubKeyBase64 string) ([]byte, error) {
...
-
-	return resBody, nil
+	switch res.StatusCode {
+	case http.StatusOK:
+		return resBody, nil
+	case http.StatusPreconditionFailed: // Return the original message from the server in case of 412
+		return nil, errors.New(strings.TrimSpace(string(resBody)))
+	default:
+		return nil, fmt.Errorf("server returned status %d", res.StatusCode)
+	}
```
jbygdell:
```suggestion
	if res.StatusCode != http.StatusOK {
		return nil, errors.New(strings.TrimSpace(string(resBody)))
	}

	return resBody, nil
```

We shouldn't treat this as a special case. It's better to always return any error string to the user.

download/download.go
```diff
@@ -89,6 +89,8 @@ type File struct {
 // Download function downloads files from the SDA by using the
 // download's service APIs
 func Download(args []string, configPath, version string) error {
+	appVersion = version
```
KarlG-nbis: Not directly related to this PR, but to me if would be cleaner if we remove the appVersion variable, and instead update the functions that uses it to take the version as a argument, eg:
``` go
func datasetCase(token, version string) error {
...
}
```
And we would pass the `version` from the `Download` invocation

pahatz (approved):
Looks good!

.github/integration/tests/40_download.sh
```diff
@@ -154,8 +154,30 @@ else
...
+# We expect this to fail because the server requires v0.2.0 or higher
+error_output=$(./sda-cli-v0.1.0 --config testing/s3cmd-download.conf download \
+    --pubkey user_key.pub.pem \
+    --dataset-id https://doi.example/ty009.sfrrss/600.45asasga \
+    --url http://localhost:8080 \
+    --outdir test-version-fail main/subfolder/dummy_data.c4gh 2>&1 || true)
+
+# Check if the error message contains the specific string returned by the server (412 logic)
+expected_msg="Your sda-cli client version is outdated, please update to at least version"
```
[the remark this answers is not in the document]
nanjiangshu: The message is like this `Error: Your sda-cli client version is outdated, please update to at least version 'v0.3.0'.`. So the string to be matched is actually correct 😄

Issue #481, opened 2024-12-09
[download] Attach a .part extension to files currently being downloaded and rename it when download completes


aaperis:
This makes it clear that a file has not finished downloading yet in case there are some errors.

PR #643, opened 2026-01-29
[download] Improving download user experience 

### LINKED ISSUE #642: [download] the --continue option is extremely slow because getBody is performed for every file
**Describe the bug**
When downloading files with the `--continue` flag, this process is extremely slow since checking the existence of files are performed after the `getBody` operation. 

Moreover, instead of performing the file existence checking in the function `downloadFile`, it should be place at the parent process to avoid the slow operation of file listing (which need to talk to the API of the download service).

**Steps to reproduce**

**Expected behavior**
- [ ] The --continue operation should skip files very fast. 

**Additional context**

nanjiangshu:
This PR closes #642 and #481

Changes made:

1. In the `downloadFile` function, the logic to check existing files have been moved to before  `getBody(uri, token, pubKeyBase64)`, since otherwise, the file content will actually be retrieved. 

2. For `fileCase`, the logic to checking existing files has been added outside of `downloadFile`, since function `getFileIDURL` is time consuming. 

3. File in downloading is added with `.part` extension and rename it to the destination when it's completed  

4. `getBody` uses streaming instead of loading everything into RAM, thus reduce the burden of the memory on the client side.

5. Improve the progress bar when streaming.  

**The result**

For example, when downloading for 10 files of the dataset `aa-Dataset-dy2tsg-9hvt2a` which are already downloaded. 
with `v0.3.1`, it takes 60 seconds
with the fix, it takes 0.01 second

**Description**

**How to test**
Tests and integration tests pass

**Manual test**
If you have access to the dataset `aa-Dataset-dy2tsg-9hvt2a`, run the following command and try to break with `ctrl-C` and rerun the command. 

```bash
./sda-cli --config s3cmd-bp-download.conf download --pubkey mykey.pub.pem --dataset-id aa-Dataset-dy2tsg-9hvt2a --url https://download.bp.nbis.se --outdir outdir  media/picostorage/default/output/encrypted/Image_lcrjOrRJbT/1.2.826.0.1.3680043.8.498.22802443563359803317234892062160314328.dcm.c4gh media/picostorage/default/output/encrypted/Image_qSuegYrtab/1.2.826.0.1.3680043.8.498.60774923466980368243827981747258811016.dcm.c4gh media/picostorage/default/output/encrypted/Image_SUgwqqoayL/1.2.826.0.1.3680043.8.498.35333794976752312564247054103896224285.dcm.c4gh media/picostorage/default/output/encrypted/Image_SUgwqqoayL/1.2.826.0.1.3680043.8.498.35333794976752312564247054103896224285.dcm.c4gh media/picostorage/default/output/encrypted/Image_hCgEQyoKVO/1.2.826.0.1.3680043.8.498.11377637496183064779528344996189878623.dcm.c4gh media/picostorage/default/output/encrypted/Image_nClErCSAvi/1.2.826.0.1.3680043.8.498.11561802173545872641119512843485577911.dcm.c4gh  --continue
```

### DIFF
--- a/list/list.go
+++ b/list/list.go
@@ -1,3 +1,4 @@
+//nolint:revive // package name conflict with stdlib is intentional
 package list
 
 import (
--- a/list/list_test.go
+++ b/list/list_test.go
@@ -1,3 +1,4 @@
+//nolint:revive // package name conflict with stdlib is intentional
 package list
 
 import (
--- a/version/version.go
+++ b/version/version.go
@@ -1,3 +1,4 @@
+//nolint:revive // package name conflict with stdlib is intentional
 package version
 
 import (
--- a/version/version_test.go
+++ b/version/version_test.go
@@ -1,3 +1,4 @@
+//nolint:revive // package name conflict with stdlib is intentional
 package version
 
 import (


download/download.go
```diff
@@ -250,6 +250,26 @@ func fileCase(args []string, token string, fileList bool) error {
...
 	for _, filePath := range files {
+		outputPath := filePath
+		if outDir != "" {
+			outputPath = outDir + "/" + filePath
+		}
+
+		outputPath = strings.TrimSuffix(outputPath, ".c4gh")
+		if pubKeyBase64 != "" {
+			outputPath += ".c4gh"
+		}
```
jbygdell:
```suggestion
```

This logic is not valid and should be removed.

jbygdell: This code always have had strange assumptions.
Combine that with a lot of spaghetti that makes it hard to make it behave properly.

download/download.go
```diff
@@ -266,59 +279,77 @@ func fileCase(args []string, token string, fileList bool) error {
...
+	// This is critical for Windows compatibility
+	if err := outFile.Close(); err != nil {
+		return fmt.Errorf("failed to close partial file %s: %v", outFile.Name(), err)
+	}
+
+	if _, err := os.Stat(filePath); err == nil {
+		if err := os.Remove(filePath); err != nil {
+			return fmt.Errorf("failed to remove existing file %s: %v", filePath, err)
+		}
+	}
```
jbygdell:
```suggestion
```

This feels like a test introduced bug.
There shouldn't be any need to delete the destination file since it wouldn't exist.

And if it did exist and the choice was to overwrite it it should be done earlier with the initial existance check.

KarlG-nbis (approved):
Looks good, great job!

download/download.go
jbygdell: The reason the test failed was because the filehandle wasn't closed properly as the error message clearly states:
```
 The process cannot access the file because it is being used by another process.
```

nanjiangshu: In the current setup of sda-cli, if you download files without the `--continue` flag, the existing files will not be skipped and thus be **overwritten**. On Windows, rename will fail if the target file exist, that's why it should be removed beforehand.

download/download.go
```diff
@@ -250,6 +251,18 @@ func fileCase(args []string, token string, fileList bool) error {
 	}
 
 	for _, filePath := range files {
+		outputPath := filepath.Join(outDir, filePath)
+
+		if continueDownload {
```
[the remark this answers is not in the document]
nanjiangshu: Yes, this is to avoid the time consuming call to the API triggered by
```
		fileIDURL, apiFilePath, err := getFileIDURL(URL, token, pubKeyBase64, datasetID, filePath)
```
Which can be quite some time if the user needs to skip thousands of files. 

But for the recursive download and `dataset` mode, query to the API is necessary and done only once, so no need to apply this logic to them.

download/download.go
nanjiangshu: That's a good point, thanks @jbygdell ! However, such change will introduce a new feature and change the behavior of the command line with user interaction, that is, it will needs users input to decide how to continue. 
Therefore, I think it is better to be address in a new issue https://github.com/NBISweden/sda-cli/issues/648. 
What do you think?

nanjiangshu: While **deleting early** is technically functional, I do think a **delete late** strategy is more robust for data integrity.

If we **delete early**, we create a window of failure: if a 2GB download is interrupted by a network drop or a crash, the user is left with no file at all because the original was already purged.

By **deleting late** (only after the .part file is fully written and closed), we ensure the user always has a valid version of the file on disk. This approach makes the file replacement more atomic and user-friendly. It is also easier to refactor if we want to implement `download by resuming from breaking point` in the future.

jbygdell: The point is that you should **never** see both the target file **AND** the part file at the same time.

```sh
filename.c4gh
filename.c4gh.part
```

The download is either in progress or have been completed.

jhagberg (commented):
Pr is in good shape 
 Just one comment in line
and maybe add the unit test @jbygdell requested

kostas-kou (approved):
Seems good to me. Good job

download/download.go
```diff
@@ -250,6 +249,18 @@ func fileCase(args []string, token string, fileList bool) error {
...
+			if _, err := os.Stat(outputPath); err == nil {
+				fmt.Printf("Skipping download to %s, file already exists\n", outputPath)
+
+				continue
+			} else if !errors.Is(err, os.ErrNotExist) {
+				return err
+			}
+		}
+
```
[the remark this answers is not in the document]
nanjiangshu: This is actually useful to avoid the time consuming step 

		fileIDURL, apiFilePath, err := getFileIDURL(URL, token, pubKeyBase64, datasetID, filePath)

especially when it iterates for thousands of files.  

see also https://github.com/NBISweden/sda-cli/pull/643#discussion_r2782374604

download/download.go
```diff
@@ -266,59 +277,75 @@ func fileCase(args []string, token string, fileList bool) error {
...
+			return fmt.Errorf("failed to remove existing file: %w", err)
+		}
 	}
 
-	fileDir := filepath.Dir(filePath)
-	err = os.MkdirAll(fileDir, 0750)
+	bodyStream, totalSize, err := getBody(uri, token, pubKeyBase64)
 	if err != nil {
-		return fmt.Errorf("failed to create directory, reason: %v", err)
+		return err
```
jbygdell:
```suggestion
		bodyStream.Close()

		return err
```

returning before closing will leak handles.

jbygdell: In the case of an error the code doesn't reach that far since the return happens earlier.

jbygdell (approved):

Issue #662, opened 2026-03-23
[sda-cli] Make a new release and update the documentation


viklund:
Make sure the documentation is up to date:
 - https://fega.nbis.se/submission/uploading-data.html
 - bp.nbis.se
 - README.md

PR #664, opened 2026-03-26
Feat: update readme

nanjiangshu:
This PR together with https://github.com/NBISweden/fega-sweden-web/pull/98 and https://github.com/NBISweden/bp-sweden-web/pull/67 fixes the issue #662 

**Description**
1. Released version v0.3.3 
2. Updated usage description and syntax 
3. Updated quotes of flags in the Go code 

**Additional notes**

The integration test will be fixed in another issue 

**How to test**

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To get help using the tool, run the following command
 
 The main functionalities of this tool are explained in the following sections.
 
-Users unfamiliar with using a command line tools in a terminal window, may find the contents of [this article](https://ftpdocs.broadcom.com/cadocs/0/CA%20ARCserve%20%20Backup%2015-ENU/Bookshelf_Files/HTML/CMD_Ref/command_line_syntax_characters.htm)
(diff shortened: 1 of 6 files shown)


KarlG-nbis (approved):

README.md
jbygdell, line 50: Should this really be here??

jbygdell (approved):

README.md
nanjiangshu: That's good point. I agree, these text should be at docs for specific project. And there are already at bp.nbis.se and fega.nbis.se. 

I will remove it.

nanjiangshu: While changing the text, I realized the project (FEGA/BP) specific description exist not only in this section but at several other sections as well. I think the purpose is to avoid people encrypt using their own public keys when uploading to FEGA/BP services. 

So probably still meaningful to keep the text here, what do you think?

jbygdell: I think any project specific things should be removed from this readme.
It will make it easier for us to move this code into the SDA repos if there only exists information on how to use the app in general terms.

nanjiangshu: I agree. But since this is not for fixing errors and instructions in the documentation but about improving consistency of the documentation in multiple repos, I've created a new issue https://github.com/NBISweden/sda-cli/issues/665 to address this issue more systematically.

Issue #666, opened 2026-04-02
Fix linter issues


kostas-kou:
The linter fails so these issues should be fixed:

```list/list_test.go:89:14: SA1019: manager.NewUploader is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
	uploader := manager.NewUploader(s3Client)
	           ^
list/list_test.go:93:15: SA1019: uploader.Upload is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
	if _, err := uploader.Upload(context.Background(), &s3.PutObjectInput{
	            ^
upload/upload.go:133:14: SA1019: manager.NewUploader is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
	uploader := manager.NewUploader(s3Client)
	           ^
upload/upload.go:224:18: SA1019: uploader.Upload is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
		result, err := uploader.Upload(ctx, &s3.PutObjectInput{
		              ^
upload/upload.go:229:14: SA1019: manager.Uploader is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
		}, func(u *manager.Uploader) {```

viklund:
Duplicate of #658.

PR #667, opened 2026-04-02
Prompt and new flags for existing files during download

### LINKED ISSUE #648: [download] do not sliently overwrite to existing file
As a user, I want I get warned when existing files exist when downloading so that choose whether to overwrite the existing file or not.

Rename `--continue` flag into `--ignore-existing`.

Cases:
 - Partial file exists delete the partial file and move on.
 - Final file exists and `--ignore-existing` flag specified => Skip download
 - Final file eixsts and `--overwrite-exiting` => Just overwite, never ask.
 - Final file exists and no `--ignore-existing` and no `--overwrite-existing` flag specified => Ask about overwriting ([Y]es/[N]o/[A]lways/ne[V]er - double check that this is how other tools name the options) 
 - It's an error in the tool to supply both `--ignore-existing` and `--overwrite-existing`.

**Acceptance criteria**
- [ ] User get proper warning message and be able to choose whether to overwrite the existing file or not

**Additional context**

see also the discussion at https://github.com/NBISweden/sda-cli/pull/643#discussion_r2781746446

kostas-kou:
This PR closes #648 .

**Description**
In this PR the following changes took place:
- `--continue` flag renamed to `--ignore-existing`
- extra flag for overwrite `--overwrite-existing`
- if none of the above flags is present then the user will be asked what to do
- if a partial and a full file is present, then the cli will delete the partial file despite the user's answer

There are also some extra changes which are not included in the issue but it was necessary to be done in order to create integration tests. Those changes are:
- Update configs and compose for running the containers after the updates in the sda repo
- Update scripts and set for running the tests without failing

There are some linter issues which are out of the scope of this PR and I have a created a separate issue for that. See ~#666~ #657

**How to test**
If one has access to a BP dataset ccan download 2-3 files and test the flags and the user prompt.


download/download.go
```diff
@@ -59,7 +61,8 @@ func init() {
 	rootcmd.AddCommand(downloadCmd)
 	downloadCmd.Flags().StringVar(&datasetID, "dataset-id", "", "Dataset ID for the file(s) to download")
 	downloadCmd.Flags().StringVar(&URL, "url", "", "The url of the download server")
-	downloadCmd.Flags().BoolVar(&continueDownload, "continue", false, "Skip existing files and continue with the rest")
+	downloadCmd.Flags().BoolVar(&continueDownload, "ignore-existing", false, "Skip existing files and continue with the rest")
```
zeidlitz: I think it makes sense to have the variable reflect the name of the flag. So in this case, either name the variable ignoreExisting, or change the flag name to be "overwrite-existing". I don't remember what we agreed on regarding the names.

download/download.go
```diff
@@ -343,6 +335,62 @@ func downloadFile(uri, token, pubKeyBase64, filePath string) error {
...
 
+func handleExistingFile(filePath string) (bool, error) {
+	if _, err := os.Stat(filePath); errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	}
+
+	// If a full file exists, any partial file is definitely garbage and should be removed
+	if _, err := os.Stat(filePath + ".part"); err == nil {
+		_ = os.Remove(filePath + ".part")
+	}
```
zeidlitz:
```suggestion
if _, err := os.Stat(partPath); err == nil {
		if err := os.Remove(partPath); err != nil {
			fmt.Printf("Warning: could not remove old partial file %s: %v\n", partPath, err)
		}
	}
```

download/download.go
```diff
@@ -343,6 +335,62 @@ func downloadFile(uri, token, pubKeyBase64, filePath string) error {
 	return nil
 }
 
+func handleExistingFile(filePath string) (bool, error) {
+	if _, err := os.Stat(filePath); errors.Is(err, os.ErrNotExist) {
+		return false, nil
+	}
+
+	// If a full file exists, any partial file is definitely garbage and should be removed
```
zeidlitz: I think it can be worth considering if this "remove old partial file" logic is even needed? Like how opinionated should the sda-cli be? Can we live without removing a .part file?

download/download.go
```diff
@@ -343,6 +335,62 @@ func downloadFile(uri, token, pubKeyBase64, filePath string) error {
...
+
+	if continueDownload {
+		fmt.Printf("Skipping download to %s, file already exists\n", filePath)
+
+		return true, nil
+	}
+
+	if !overwriteExisting {
+		if sessionOverwrite == helpers.OverwriteNever {
+			fmt.Printf("Skipping download to %s, files already exists\n", filePath)
```
zeidlitz:
```suggestion
			fmt.Printf("Skipping download to %s, file already exist\n", filePath)
```

download/download.go
```diff
@@ -343,6 +335,62 @@ func downloadFile(uri, token, pubKeyBase64, filePath string) error {
...
+			}
+
+			switch choice {
+			case helpers.OverwriteAlways:
+				sessionOverwrite = helpers.OverwriteAlways
+			case helpers.OverwriteNever:
+				sessionOverwrite = helpers.OverwriteNever
+				fmt.Printf("Skipping download to %s, file already exists\n", filePath)
+
+				return true, nil
```
zeidlitz:
```suggestion
				fallthrough
```

download/download.go
```diff
@@ -343,6 +335,62 @@ func downloadFile(uri, token, pubKeyBase64, filePath string) error {
 	return nil
 }
 
+func handleExistingFile(filePath string) (bool, error) {
```
zeidlitz, line 356: would it make sense to pass all the needed variables in the function. So the signature would be


```suggestion
func handleExistingFile(filePath, continueDownload, overwriteExisting, sessionOverwrite string) (bool, error) {
```

Just to be extra explicit, and could make this easier to test I guess?

helpers/helpers.go
```diff
@@ -71,6 +72,46 @@ func PromptPassword(message string) (password string, err error) {
...
+			return OverwriteNo, nil
+		case "a", "always":
+			return OverwriteAlways, nil
+		case "v", "never":
+			return OverwriteNever, nil
+		default:
+			fmt.Println("Invalid input, please try again.")
+		}
+	}
+}
```
zeidlitz: Would something like this make sense?


```suggestion
func PromptOverwrite(filename string) (OverwriteChoice, error) {
	reader := bufio.NewReader(os.Stdin)
	
	inputMap := map[string]OverwriteChoice{
		"y": OverwriteYes, "yes": OverwriteYes,
		"n": OverwriteNo, "no": OverwriteNo,
		"a": OverwriteAlways, "always": OverwriteAlways,
		"v": OverwriteNever, "never": OverwriteNever,
	}

	for {
		fmt.Printf("File %s exists. Overwrite? [y/n/a/v]: ", filename)
		input, err := reader.ReadString('\n')
		if err != nil {
			return OverwriteNone, err
		}

		cleanInput := strings.ToLower(strings.TrimSpace(input))
		if choice, ok := inputMap[cleanInput]; ok {
			return choice, nil
		}

		fmt.Println("Invalid input. Please use (y)es, (n)o, (a)lways, or (v)ever.")
	}
}
```

To make it a bit more compact and readable?

zeidlitz (approved):
LGTM, added some thoughts I think can be worth considering but nothing major

download/download.go
```diff
@@ -243,18 +251,6 @@ func fileCase(args []string, token string, fileList bool) error {
 	}
 
 	for _, filePath := range files {
-		outputPath := filepath.Join(outDir, filePath)
```
nanjiangshu, line 246: This logic seems a bit redundant but needs to be here. Otherwise, the following function call `getFileIDURL` will further call the function`GetFilesInfo` (which call the download API) for every file. 

This will not only be a burden for the API, but also making the `ignore-existing` process slow -- takes about half a second per file.

nanjiangshu (changes requested):
Looks good and all combinations of the flags works as expected. However, the `--ignore-existing` process is now slow since it will call the download API for every file. 

The fix would be either bring back the seemly redundant code (see the inline comment) or make a more proper refactoring so that the `GetFilesInfo` is called only once

download/download.go
kostas-kou: You are right, renamed it

download/download.go
kostas-kou: By default in the package when the cli downloads a file and partial file is present it will be deleted and replaced by a the full file. We do that as an extra thing because there was a corner case were both full and partial files were present so whatever flag you use the partial file always be there. Maybe @nanjiangshu can correct me if I am wrong

download/download.go
kostas-kou: Done

download/download.go
```diff
@@ -343,6 +335,64 @@ func downloadFile(uri, token, pubKeyBase64, filePath string) error {
...
+				fmt.Printf("Skipping download to %s, file already exists\n", filePath)
+
+				return true, nil
+			case helpers.OverwriteYes:
+				// Proceed to remove and download
+			default:
+				return false, fmt.Errorf("unknown overwrite choice: %v", choice)
+			}
+		}
+	}
```
jbygdell, line 389: This is logically heavy and should probably be simplified with less nesting.

For example `overwriteExisting` and `helpers.OverwriteAlways`  would give the same end result.
The same goes for `ignoreExisting` and `helpers.OverwriteNever`. 


If something like this where to be used the `sessionOverwrite` variable can be removed and the nesting simplified.
```
case helpers.OverwriteAlways:
	overwriteExisting = true
case helpers.OverwriteNever:
	ignoreExisting = true
```

download/download.go
kostas-kou: Good point about testing. Is just a personal preference to avoid many inputs in a function if I can avoid it since all of them are global variables and in the test you can pass them as downloadCmd.Flag
There is no explicit test for this function since there are unit tests for cases so the downloadFile function is used.
I mean I do not have strong preferences so if you think that this will improve the clarity I can change it

helpers/helpers.go
kostas-kou: This looks awesome.

download/download.go
kostas-kou: You are right. It is there now

viklund:
There already was an issue for the linter issues, #657 created by @pahatz.

download/download.go
kostas-kou: Good point. I have made changes in this function as well as in the unit tests.

download/download.go
nanjiangshu: That's correct, though real corner case, but good to tidy up :)

nanjiangshu (approved):
Great work!

jbygdell (approved):

Issue #663, opened 2026-03-24
feat: support SDA download API v2 endpoints


jhagberg:
### Context

The SDA download service is getting a v2 API with different endpoints, response
shapes, and header conventions. The plan is to run v1 and v2 in parallel during
migration. The CLI should support both API versions.

### Current CLI → v1 API mapping

| CLI function | v1 endpoint | Header |
|---|---|---|
| `GetDatasets` | `GET /metadata/datasets` | — |
| `GetFilesInfo` | `GET /metadata/datasets/{id}/files` | — |
| Download file | `GET /s3/{datasetId}/{fileName}` | `Client-Public-Key` |

Response shapes: `GetDatasets` returns `[]string`, `GetFilesInfo` returns
`[]File` (flat array with `fileId`, `filePath`, `fileSize`,
`decryptedFileSize`, etc.).

### v2 API differences

| Aspect | v1 | v2 |
|---|---|---|
| List datasets | `GET /metadata/datasets` → `[]string` | `GET /datasets` → `{ datasets: [...], nextPageToken }` |
| List files | `GET /metadata/datasets/{id}/files` → `[]File` | `GET /datasets/{id}/files` → `{ files: [...], nextPageToken }` |
| Download | `GET /s3/{datasetId}/{fileName}` | `GET /files/{fileId}` |
| Public key header | `Client-Public-Key` | `X-C4GH-Public-Key` |
| Error format | Mixed | RFC 9457 Problem Details (JSON) |
| Pagination | None | HMAC-signed `nextPageToken` (keyset pagination) |
| File ID download | Via path (`/s3/dataset/filename`) | Via stable ID (`/files/{fileId}`) |
| Split endpoints | N/A | `/files/{fileId}/header`, `/files/{fileId}/content` |

### Proposed changes

1. **API version detection or flag** — add `--api-version v1|v2` flag
   (default `v1`). In the future when it works we switch default to be `v2`.

2. **Paginated list endpoints** — `GetDatasets` and `GetFilesInfo` must follow
   `nextPageToken` to collect all results. v1 returns everything in one response.

3. **Download by file ID** — v2 uses `GET /files/{fileId}` instead of
   constructing `/s3/{datasetId}/{fileName}`. The CLI's `getFileIDURL` function
   and `downloadFile` URL construction need updating.

4. **Header rename** — `Client-Public-Key` → `X-C4GH-Public-Key`.

[...]

PR #679, opened 2026-04-23
feat(apiclient): Client interface + V1Client wrapper

### LINKED ISSUE #674: feat(download): introduce apiclient abstraction behind --api-version flag
Part of #663.

As an sda-cli maintainer, I want a `Client` interface in front of the download API so later PRs can plug in a v2 implementation without touching download or list command code.

**Acceptance criteria**
- [ ] New package `apiclient/` contains: `Client` interface (`ListDatasets`, `ListFiles`, `DatasetInfo`), types (`File`, `Checksum`, `DatasetInfo`, `ListFilesOptions`), `ErrNotSupportedOnV1`, and `New(cfg, version)` factory
- [ ] `V1Client` in `apiclient/v1.go`: v1 HTTP logic moved from `download/download.go` (`GetDatasets`, `GetFilesInfo`, `getBody`, persistent cookie jar) with identical wire behavior
- [ ] `download.GetDatasets` / `download.GetFilesInfo` keep working as thin delegates, `download.File` becomes a type alias of `apiclient.File` (both go away in #677)
- [ ] `--api-version v1|v2` flag on `download` and `list` (default `v1`); `v2` errors with "not yet implemented"
- [ ] `datasetCase` / `recursiveCase` / `fileCase` / `getFileIDURL` take an `apiclient.Client` and call it for listing; `list.Datasets` / `list.datasetFiles` build the client via `apiclient.New`
- [ ] Cookie file stays at `${UserCacheDir}/sda-cli/sda_cookie` (session compat)
- [ ] All existing `go test ./...` tests pass unchanged; `TestInvalidUrl` still gets `"failed to get files, reason:"`

**Additional context**
Out of scope here: `DownloadFile` on the interface. It joins in #677 when v2 download lands — adding it now forces a progress-bar refactor that does not belong in an abstraction PR.

jhagberg:
Closes #674. First of six PRs for #663 (SDA download v2 support); #681, #685, #686, #687 stack on top.

**Description**
Introduces `apiclient.Client` as a version-agnostic interface for the download HTTP surface, with `V1Client` wrapping the existing v1 code and `V2Client` left as a stub for #681+. All `list` and `download` list-family calls route through the client; file transfer stays on the legacy path until #686.

A new `--api-version v1|v2` flag (default `v1`) picks the implementation. `v2` errors with "not yet implemented" until #681 wires up `V2Client.ListDatasets`.

V1 wire behavior is preserved: same headers (`SDA-Client-Version`, `Authorization`, `Content-Type`, optional `Client-Public-Key`), same cookie path (`${UserCacheDir}/sda-cli/sda_cookie`), same `"failed to get files, reason:"` error prefix asserted by `TestInvalidUrl`, same HTTP 412 handling. `DownloadFile` joins the interface in #686 to avoid coupling this PR to the progress-bar refactor.

Updated from review (nanjiangshu): `Config.Version` → `Config.ClientVersion`, the `version` parameter on `New`/`ValidateVersion` → `apiVersion`, and `File.DecryptedFileSize` widened to `int64` (matches sda-download's wire type; avoids truncation above 2 GiB on 32-bit clients).

**How to test**
- `go test ./...` (169 pass), `go vet ./...`, `golangci-lint run` — all clean
- `sda-cli list --datasets` and `sda-cli download --dataset-id X ...` work as before on `--api-version v1`
- `--api-version v2` errors early with "not yet implemented"
- `--api-version v9` errors with `unsupported --api-version "v9" (v1 or v2)`

## Test plan
- [x] `go test ./...`: 169 pass
- [x] `go vet ./...`: clean
- [x] `gofmt -l .`: clean
- [x] Manual: v1 path unchanged, v2 errors early, cookie path unchanged
- [ ] CI: golint, test (ubuntu + windows), integration (s3), codeql-analysis

### DIFF
--- a/apiclient/apiclient.go
+++ b/apiclient/apiclient.go
@@ -1,0 +1,75 @@
+package apiclient
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"go.nhat.io/cookiejar"
+)
+
+// Config holds per-call configuration.
+type Config struct {
+	BaseURL string // e.g. "https://download.example.org"
+	Token   string // bearer token (raw, no "Bearer " prefix)
+	Version string // sda-cli version; sent as SDA-Client-Version header
+}
+
+// Client is the SDA download API abstraction for list-family operations.
+// The DownloadFile method joins this interface in #677 alongside v2
+// download implementation.
+type Client interface {
+	ListDatasets(ctx context.Context) ([]string, error)
+	ListFiles(ctx context.Context, datasetID string, opts ListFilesOptions) ([]File, error)
+	DatasetInfo(ctx context.Context, datasetID string) (DatasetInfo, error)
+}
+
+// Option customises the Client returned by New. Options only affect the
+// versions they apply to; unknown options are silently ignored by
+// versions that do not honour them (WithV1CookieJar is v1-only).
+type Option func(*clientOpts)
+
+type clientOpts struct {
+	v1CookieJar *cookiejar.PersistentJar
+}
+
+// WithV1CookieJar hands V1Client an externally-managed persistent cookie
+// jar instead of letting it lazy-init its own. Required when the caller
+// (e.g. download/download.go) also runs the legacy downloadFile path so
+// metadata listing and /s3 transfer share the same in-memory jar and
+// avoid clobbering each other via AutoSync to the shared on-disk file.
+func WithV1CookieJar(jar *cookiejar.PersistentJar) Option {
+	return func(o *clientOpts) { o.v1CookieJar = jar }
+}
+
+// ValidateVersion returns the same error shape as New for a given
+// version string without constructing a client. Callers can use it to
+// fail fast on unsupported versions before doing other setup work
+// (e.g. cookie-jar init) that is wasted when the command is going to
+// error out anyway.
+func ValidateVersion(version string) error {
+	switch version {
+	case "v1":
+		return nil
+	case "v2":
+		return errors.New("--api-version v2 is not yet implemented; see #663 for progress")
+	default:
+		return fmt.Errorf("unsupported --api-version %q (v1 or v2)", version)
+	}
+}
+
+// New returns a Client for the requested version.
+// Today accepts "v1" only; "v2" errors. Extended in #675 to return a V2Client.
+func New(cfg Config, version string, opts ...Option) (Client, error) {
+	if err := ValidateVersion(version); err != nil {
+		return nil, err
+	}
+
+	var o clientOpts
+	for _, opt := range opts {
+		opt(&o)
+	}
+	// ValidateVersion above guarantees version is "v1" (the only
+	// branch that returns a client in this implementation).
+	return NewV1Client(cfg, o.v1CookieJar), nil
+}
--- a/apiclient/types.go
+++ b/apiclient/types.go
@@ -1,0 +1,45 @@
+// Package apiclient defines the SDA download API client interface and
+// shared types. v1 and v2 implementations live in v1.go and v2.go.
+package apiclient
+
+// File is the shared metadata type for a dataset file. Fields match v1's
+// wire shape. v2 has additional fields (checksums array, downloadUrl)
+// that are handled inside V2Client's conversion layer and surfaced
+// through this same type where applicable (see #675/#677).
+type File struct {
+	FileID                    string `json:"fileId"`
+	DisplayFileName           string `json:"displayFileName"`
+	FilePath                  string `json:"filePath"`
+	DecryptedFileSize         int    `json:"decryptedFileSize"`
+	DecryptedFileChecksum     string `json:"decryptedFileChecksum"`
+	DecryptedFileChecksumType string `json:"decryptedFileChecksumType"`
+}
+
+// Checksum is the v2 checksum element. Included for completeness; v1
+// does not populate it. Used by v2 types in #675.
+type Checksum struct {
+	Type     string `json:"type"`
+	Checksum string `json:"checksum"`
+}
+
+// DatasetInfo is the v2 /datasets/{id} metadata response. v1 has no
+// equivalent; V1Client.DatasetInfo returns ErrNotSupportedOnV1.
+type DatasetInfo struct {
+	DatasetID string `json:"datasetId"`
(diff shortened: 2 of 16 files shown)


apiclient/apiclient.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+
+	"go.nhat.io/cookiejar"
+)
+
+// Config holds per-call configuration.
+type Config struct {
+	BaseURL string // e.g. "https://download.example.org"
+	Token   string // bearer token (raw, no "Bearer " prefix)
+	Version string // sda-cli version; sent as SDA-Client-Version header
```
nanjiangshu: It is probably better to rename this as `ClientVersion` to reduce name confusion.

```suggestion
	ClientVersion string // sda-cli version (e.g. v1.2.3); sent as SDA-Client-Version header
```

apiclient/apiclient.go
```diff
@@ -0,0 +1,75 @@
...
+	case "v2":
+		return errors.New("--api-version v2 is not yet implemented; see #663 for progress")
+	default:
+		return fmt.Errorf("unsupported --api-version %q (v1 or v2)", version)
+	}
+}
+
+// New returns a Client for the requested version.
+// Today accepts "v1" only; "v2" errors. Extended in #675 to return a V2Client.
+func New(cfg Config, version string, opts ...Option) (Client, error) {
```
nanjiangshu: And here to rename `version` to `apiVersion` to clearly distinguish it from the clientVersion

nanjiangshu (commented):
Works great when I tested. I just left a minor comment regarding potential naming confusion for `version`. Otherwise, it looks good.

kostas-kou (approved):
Tested it and is working. Good job

apiclient/types.go
```diff
@@ -0,0 +1,45 @@
...
+
+// File is the shared metadata type for a dataset file. Fields match v1's
+// wire shape. v2 has additional fields (checksums array, downloadUrl)
+// that are handled inside V2Client's conversion layer and surfaced
+// through this same type where applicable (see #675/#677).
+type File struct {
+	FileID                    string `json:"fileId"`
+	DisplayFileName           string `json:"displayFileName"`
+	FilePath                  string `json:"filePath"`
+	DecryptedFileSize         int    `json:"decryptedFileSize"`
```
nanjiangshu: Why not use `int64`. This is also how it is defined in sda-download (v1) https://github.com/neicnordic/sensitive-data-archive/blob/main/sda-download/internal/database/database.go#L34

Although the definition data type for filesizes are not consistent in sda-download.

But from the client side, it is better to define as `int64` than `int` since we can not guarantee the system on the client side.

apiclient/apiclient.go
jhagberg: Good point, will rename `Config.Version` to `Config.ClientVersion` and take the doc tweak from your suggestion. The `version` parameter rename below goes in the same amend.

apiclient/apiclient.go
jhagberg: Yes, will do. `apiVersion` matches the `--api-version` flag name too, which is handy.

apiclient/types.go
jhagberg: Good catch, thanks for flagging it! We caught the same thing in an earlier internal review pass. sda-download is `int64` server-side, and on a 32-bit client the current `int` would truncate above 2 GiB.

Plan is to widen the field in the amend. `list.go` needs the same widening (`datasetSize` and `formatFileSizeOutput`), and the `int(f.DecryptedSize)` cast that PR #685 currently has in `apiclient/v2_types.go` falls out as a follow-up.

apiclient/apiclient.go
```diff
@@ -0,0 +1,75 @@
+package apiclient
```
KarlG-nbis: would it make more sense to name this package "downloadclient", etc to be more specific in which client this package is an interface for? Just to reduce risk of mix up with the "sda-api", etc?

downloadclient/downloadclient.go
```diff
@@ -0,0 +1,75 @@
...
+	case "v2":
+		return errors.New("--api-version v2 is not yet implemented; see #663 for progress")
+	default:
+		return fmt.Errorf("unsupported --api-version %q (v1 or v2)", apiVersion)
+	}
+}
+
+// New returns a Client for the requested apiVersion.
+// Today accepts "v1" only; "v2" errors. Extended in #675 to return a V2Client.
+func New(cfg Config, apiVersion string, opts ...Option) (Client, error) {
```
KarlG-nbis, line 66: Would we also want validation of the `cfg Config` values here, or are we just assuming that the caller has validated those already?

nanjiangshu (approved):
Looks good to me now! Well done!

apiclient/apiclient.go
jhagberg: Good point. Looking at it more, I think the ambiguity with the inbox sda-api will only get worse as more list and download paths land in this package, so I folded the rename into this PR rather than chasing it as a follow-up.

The cascade onto the four stacked PRs (#681, #685, #686, #687) was less painful than expected. The package moves from `apiclient/` to `downloadclient/`, `apiclient.go` becomes `downloadclient.go` (git's rename detection picks that up, so it shows as a rename rather than delete+add), and the rest of the files keep their names. Each stacked branch is rebased on top and verified green: `go test`, `go vet`, `gofmt`, and `golangci-lint --build-tags=integration` all clean :)

downloadclient/downloadclient.go
jhagberg: Done, fail-fast in `New()`. Added a small `Config.validate()` helper that catches an empty `BaseURL`, a `BaseURL` that fails `url.ParseRequestURI`, and an empty `Token`. `ClientVersion` stays optional because it is only emitted as a header.

Worth flagging at review: the two `TestListDataset...NoUrl` cases in `list/list_test.go` had to swap their expected error string from the old downstream `"invalid base URL"` (which came from the per-call validation in `V1Client.ListDatasets`) to the new `"Config.BaseURL is required"` because the construction-time check fires earlier. New tests in `downloadclient_test.go` cover the four error shapes.

KarlG-nbis (approved):

PR #681, opened 2026-04-23
feat(download): v2 CI harness + minimal ListDatasets

### LINKED ISSUE #674: feat(download): introduce apiclient abstraction behind --api-version flag

### LINKED ISSUE #675: feat(download): v2 API CI harness + minimal /datasets listing
Part of #663. Depends on #674.

As an sda-cli developer, I want CI to run each v2 PR against the real v2 dev stack. Contract mismatches should show up in review, not at release.

**Acceptance criteria**
- [ ] `apiclient/v2.go`: `V2Client` implementing the `Client` interface. `ListDatasets` does a single-page `GET /datasets`. If the response has a non-null `nextPageToken`, the call returns an error so we do not silently truncate (the paginator lands in #SI-3)
- [ ] `apiclient/v2_types.go`: v2 wire types (`v2File`, `datasetListResponse`, `fileListResponse`, `datasetInfoResponse`). `v2File.toFile()` maps the `checksums` array to the legacy scalar fields (prefer `sha256`)
- [ ] `apiclient.New("v2", cfg)` returns a real `V2Client`
- [ ] `.github/workflows/integration-v2.yml`: checks out `neicnordic/sensitive-data-archive` at a pinned commit, runs `make dev-download-v2-up`, waits for readiness, fetches a dev token from mockauth, runs `go test -tags=integration ./...`
- [ ] `docs/v2-dev-stack-pin.md`: records the pinned commit and how to bump it
- [ ] `TestV2_ListDatasets_Smoke` asserts the seeded `EGAD00000000001` dataset is returned

**Additional context**
Dev stack lives at `dev-tools/download-v2-dev/` (added in [neicnordic/sensitive-data-archive#2368](https://github.com/neicnordic/sensitive-data-archive/pull/2368)).

jhagberg:
Closes #675. Stacks on #679. Part of umbrella issue #663.

**Description**
Wires `V2Client.ListDatasets` (single-page) against the real v2 server, lands the v2 wire types (`v2File`, response envelopes, `datasetInfoResponse`) that #685 and #686 consume, and adds a CI workflow that boots a pinned `neicnordic/sensitive-data-archive` dev stack for integration tests. `ListFiles` and `DatasetInfo` remain stubs until #685. Also restores the legacy `"failed to get X, reason: ..."` error-wrap prefix for the `list` command (accidentally dropped when the apiclient abstraction landed in #679).

Updated from review (nanjiangshu): `V2Client.ListDatasets` uses `url.JoinPath` instead of string concatenation, `getJSON` now sends `SDA-Client-Version` alongside `User-Agent` so v1-style log pipelines still see the version on v2, and the `int(f.DecryptedSize)` cast in `v2File.toFile()` drops out following the `File.DecryptedFileSize` widening to `int64` in #679. The HTTP-client timeout question from review is being discussed inline — `Client.Timeout` would cut multi-GB downloads in #686, so the counter-suggestion is `Transport.ResponseHeaderTimeout`.

**How to test**
- `go build ./...`, `go vet ./...`, `go test ./...` (174 pass) — clean.
- Integration: `go test -tags=integration ./...` against a booted dev stack — `TestV2_ListDatasets_Smoke` passes.
- `sda-cli list --datasets --api-version v2` prints the seeded dataset IDs.
- `sda-cli list --dataset X --api-version v2` errors with the stub message until #685.
- CI job `integration-v2` exercises the full setup on GitHub Actions.

## Test plan
- [x] `go test ./...`: 174 pass
- [x] `go vet ./...`: clean
- [x] `gofmt -l .`: clean
- [x] Integration: `TestV2_ListDatasets_Smoke` against the live dev stack
- [x] Manual: v2 `--datasets` prints `EGAD00000000001`; `--dataset X --api-version v2` returns stub error; `--api-version v9` rejected

### DIFF
--- a/apiclient/v2.go
+++ b/apiclient/v2.go
@@ -43,7 +43,7 @@ func (c *V2Client) ListDatasets(ctx context.Context) ([]string, error) {
 		return nil, fmt.Errorf("failed to decode /datasets response: %w", err)
 	}
 	if resp.NextPageToken != nil && *resp.NextPageToken != "" {
-		return nil, errors.New("pagination not yet implemented (coming in PR #3 of issue #663); v2 response had nextPageToken but paginate helper is not wired yet")
+		return nil, errors.New("pagination not yet implemented (coming in #676)")
 	}
 
 	return resp.Datasets, nil
--- a/apiclient/v2.go
+++ b/apiclient/v2.go
@@ -1,0 +1,91 @@
+package apiclient
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// V2Client talks to the v2 SDA download API
+// (GET /datasets, /datasets/{id}/files, /files/{id}, etc.).
+// Methods fill in across #675, #676, #677. Until then, unimplemented
+// methods return a clear "not implemented until #N" error.
+type V2Client struct {
+	cfg  Config
+	http *http.Client
+}
+
+// NewV2Client constructs a V2Client. HTTP client is a plain net/http.Client
+// (no cookie jar — v2 is stateless bearer-token auth, no Location-based
+// redirects requiring cookies).
+func NewV2Client(cfg Config) *V2Client {
+	return &V2Client{
+		cfg:  cfg,
+		http: &http.Client{},
+	}
+}
+
+// ListDatasets implements Client. Single-page only here; pagination via the
+// paginate[T] helper arrives in #676. Returning an explicit error when
+// nextPageToken != null prevents silently truncating results.
+func (c *V2Client) ListDatasets(ctx context.Context) ([]string, error) {
+	body, err := c.getJSON(ctx, c.cfg.BaseURL+"/datasets")
+	if err != nil {
+		return nil, err
+	}
+	defer body.Close() //nolint:errcheck
+
+	var resp datasetListResponse
+	if err := json.NewDecoder(body).Decode(&resp); err != nil {
+		return nil, fmt.Errorf("failed to decode /datasets response: %w", err)
+	}
+	if resp.NextPageToken != nil && *resp.NextPageToken != "" {
+		return nil, errors.New("pagination not yet implemented (coming in PR #3 of issue #663); v2 response had nextPageToken but paginate helper is not wired yet")
+	}
+
+	return resp.Datasets, nil
+}
+
+// ListFiles implements Client. Not implemented until #676.
+func (c *V2Client) ListFiles(_ context.Context, _ string, _ ListFilesOptions) ([]File, error) {
+	return nil, errors.New("V2Client.ListFiles not implemented until #676")
+}
+
+// DatasetInfo implements Client. Not implemented until #676.
+func (c *V2Client) DatasetInfo(_ context.Context, _ string) (DatasetInfo, error) {
+	return DatasetInfo{}, errors.New("V2Client.DatasetInfo not implemented until #676")
+}
+
+// getJSON performs an authenticated GET returning the response body.
+// #678 replaces error wrapping with RFC 9457 Problem Details parsing.
+func (c *V2Client) getJSON(ctx context.Context, reqURL string) (io.ReadCloser, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("build request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+c.cfg.Token)
+	req.Header.Set("Accept", "application/json")
+	if c.cfg.Version != "" {
+		req.Header.Set("User-Agent", "sda-cli/"+c.cfg.Version)
+	}
+
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("http request: %w", err)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		b, _ := io.ReadAll(resp.Body)
+		_ = resp.Body.Close()
+		body := string(b)
+		if len(body) > 200 {
+			body = body[:200]
+		}
+
+		return nil, fmt.Errorf("server returned status %d: %s", resp.StatusCode, body)
+	}
+
+	return resp.Body, nil
+}
--- a/apiclient/v2_types.go
+++ b/apiclient/v2_types.go
@@ -1,0 +1,70 @@
+package apiclient
+
+// v2File is the v2 server's FileInfo wire shape. Differs from apiclient.File:
+//   - Scalar DecryptedFileSize → DecryptedSize (int64 here vs int in v1)
+//   - DecryptedFileChecksum/Type → Checksums array
+//   - New DownloadURL (server-provided; clients must not construct /files/{id})
+//   - Drops v1-only fields (DisplayFileName, etc.)
+//
+// Conversion to the shared File type happens at the V2Client boundary.
+type v2File struct {
+	FileID        string     `json:"fileId"`
+	FilePath      string     `json:"filePath"`
+	Size          int64      `json:"size"`
+	DecryptedSize int64      `json:"decryptedSize"`
+	Checksums     []Checksum `json:"checksums"`
+	DownloadURL   string     `json:"downloadUrl"`
+}
+
+// datasetListResponse is the v2 response for GET /datasets.
+// nextPageToken is nullable per swagger — use a pointer.
+type datasetListResponse struct {
+	Datasets      []string `json:"datasets"`
+	NextPageToken *string  `json:"nextPageToken"`
+}
+
+// fileListResponse is the v2 response for GET /datasets/{id}/files.
+type fileListResponse struct {
+	Files         []v2File `json:"files"`
+	NextPageToken *string  `json:"nextPageToken"`
+}
+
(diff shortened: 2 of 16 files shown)


downloadclient/v2.go
```diff
@@ -0,0 +1,91 @@
...
+	http *http.Client
+}
+
+// NewV2Client constructs a V2Client. HTTP client is a plain net/http.Client
+// (no cookie jar — v2 is stateless bearer-token auth, no Location-based
+// redirects requiring cookies).
+func NewV2Client(cfg Config) *V2Client {
+	return &V2Client{
+		cfg:  cfg,
+		http: &http.Client{},
```
nanjiangshu, line 34: It is better to set a timeout on the HTTP client here? Without the timeout, requests may hang indefinitely. And this can be a problem especially when sda-cli is called repeatedly called by a script. 


```suggestion
		http: &http.Client{Timeout: 30 * time.Second},
```

apiclient/v2.go
```diff
@@ -0,0 +1,91 @@
...
+		cfg:  cfg,
+		http: &http.Client{},
+	}
+}
+
+// ListDatasets implements Client. Single-page only here; pagination via the
+// paginate[T] helper arrives in #676. Returning an explicit error when
+// nextPageToken != null prevents silently truncating results.
+func (c *V2Client) ListDatasets(ctx context.Context) ([]string, error) {
+	body, err := c.getJSON(ctx, c.cfg.BaseURL+"/datasets")
```
nanjiangshu: Better to use `url.JoinPath(c.cfg.BaseURL, "datasets")` for url concatenation.

apiclient/v2.go
```diff
@@ -0,0 +1,91 @@
...
+// #678 replaces error wrapping with RFC 9457 Problem Details parsing.
+func (c *V2Client) getJSON(ctx context.Context, reqURL string) (io.ReadCloser, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("build request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+c.cfg.Token)
+	req.Header.Set("Accept", "application/json")
+	if c.cfg.Version != "" {
+		req.Header.Set("User-Agent", "sda-cli/"+c.cfg.Version)
```
nanjiangshu: I think we use `SDA-Client-Version` to pass the version of sda-cli client, have I missed something?

apiclient/v2_types.go
```diff
@@ -0,0 +1,70 @@
...
+
+// toFile converts a v2File into the shared apiclient.File.
+// Maps the Checksums array to the legacy scalar fields:
+// prefer sha256 if present, else first entry.
+// DisplayFileName is not populated (v2 doesn't return it).
+func (f v2File) toFile() File {
+	out := File{
+		FileID:            f.FileID,
+		FilePath:          f.FilePath,
+		DecryptedFileSize: int(f.DecryptedSize),
```
nanjiangshu: If we define `DecryptedSize` also as `int64`, then there's no need to cast `int64` to `int`, which may potentially cause the overflow problem on a 32-bit system.

downloadclient/v2.go
jhagberg: Good point on the indefinite hang in scripts. The catch is #686 sends file downloads through this same `http.Client`, and `Client.Timeout` covers the whole request including the body read, so a multi-GB download on a slow link would be cut at 30s. Would `Transport.ResponseHeaderTimeout` (bounds time-to-first-byte without capping streaming reads) work instead? Happy to put that in now if you'd prefer it over per-call `context.WithTimeout` at the list sites later.

apiclient/v2.go
jhagberg: Good catch, switching to `url.JoinPath(c.cfg.BaseURL, "datasets")`. Threads an error through since it returns one, but worth it for the trailing-slash handling.

apiclient/v2.go
jhagberg: You haven't missed anything, v2 sends `User-Agent: sda-cli/<version>` here. The User-Agent route felt closer to standard HTTP convention for a stateless bearer-token API, but you're right that anything downstream set up around `SDA-Client-Version` would go blind on v2. Added `SDA-Client-Version` alongside the User-Agent in the amend so v2 traffic stays visible to either log pipeline.

apiclient/v2_types.go
jhagberg: Yes, exactly. The same concern came up on #679 from your review there (`File.DecryptedFileSize` itself). I've amended #679 to widen the public field to `int64`, so after rebasing this PR the cast just drops out (`DecryptedFileSize: f.DecryptedSize`, both `int64`). Visible in the rebased diff.

downloadclient/v2.go
nanjiangshu: It's good enough if timeout is already covered by the parent process.

nanjiangshu (approved):
Looks good to me, well done!

downloadclient/v2.go
jhagberg: Cool, thanks for the back-and-forth on this. Agreed that for a one-shot CLI the parent-process signal handling is good enough, so leaving the timeout out. Will revisit if we ever wrap this client into a long-running process :)

list/list.go
```diff
@@ -120,7 +120,16 @@ func datasetFiles(token string, url string, dataset string, bytesFormat bool) er
 
 	files, err := client.ListFiles(context.Background(), dataset, downloadclient.ListFilesOptions{})
 	if err != nil {
-		return err
+		// URL-validation errors are returned unwrapped so TestListDatasetNoUrl
+		// still sees the bare "invalid base URL" string; transport / parse /
+		// HTTP errors get the legacy "failed to get files, reason: ..."
+		// prefix that callers of the previous download.GetFilesInfo shim
+		// expected before downloadclient was introduced.
+		if err.Error() == "invalid base URL" {
```
zeidlitz: I'm not that knowledgeable about error comparisons in Go, but is there a reason to not use errors.Is here?

downloadclient/v2.go
```diff
@@ -0,0 +1,97 @@
...
+
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("http request: %w", err)
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		b, _ := io.ReadAll(resp.Body)
+		_ = resp.Body.Close()
+		body := string(b)
+		if len(body) > 200 {
```
zeidlitz: could we clarify exactly what this 200 number is? Just extract it as a variable name for extra clarity?

eg something like

`const maxErrorBodyBytes = 200`

zeidlitz (approved):
Looks pretty straight forward to me. Left a few questions for curiosity.

list/list.go
jhagberg: Good question. The block is checking against a string error from `downloadclient.New`'s URL validation, so `errors.Is` would need a sentinel (e.g. `var ErrInvalidBaseURL = errors.New(...)`) on the downloadclient side first.

Honest answer though: this block is dead code after #679's review-amend added fail-fast validation in `New()`. The new error is `"Config.BaseURL is required"` (not `"invalid base URL"`), and it fires before `ListDatasets` is ever called, so the `if` branch never matches. There's an identical dead block in `Datasets()` further down. I think the cleanest move is to delete both rather than wire `errors.Is` for an unreachable path. Will clean them up.

downloadclient/v2.go
jhagberg: Good shout. Will extract `const maxErrorBodyBytes = 200` at the package level and use it at both the `len(body) > N` check and the `body[:N]` truncation. The same constant becomes useful in #685's chore commit (caps the body read with `io.LimitReader(resp.Body, maxErrorBodyBytes+1)`), so it propagates naturally up the stack.

zeidlitz (approved):

PR #684, opened 2026-04-24
fix(encrypt): stabilize TestStream by dropping scheduling-dependent assertions

### LINKED ISSUE #683: TestEncryptTestSuite/TestStream is flaky on windows-latest (race in encrypt.Stream)
**Describe the bug**

`TestEncryptTestSuite/TestStream` in `encrypt/encrypt_test.go` fails intermittently on `windows-latest`. A re-run of the same commit passes.

First hit on PR #682, run https://github.com/NBISweden/sda-cli/actions/runs/24845578520/job/72731548937.

**Steps to reproduce**

1. Push any commit that includes `encrypt/encrypt_test.go` and run the `Go` workflow.
2. Wait for the `Test (windows-latest)` job. It fails occasionally; a re-run passes.

Mechanism: `encrypt.Stream()` (`encrypt/encrypt.go:570-616`) sets up a `TeeReader` that updates `UnencryptedMD5` on every read, then starts a goroutine that runs `io.Copy(crypt4GHWriter, unencryptedReader)`. The goroutine updates `UnencryptedMD5` as it copies.

The test at `encrypt/encrypt_test.go:266` asserts `UnencryptedMD5` is the MD5 of the empty string right after `Stream()` returns. On Linux the assertion usually runs before the goroutine has a chance to do anything. On Windows the scheduling sometimes goes the other way. For a 7-byte file the goroutine finishes the whole copy first.

**Expected behavior**

- [ ] `TestStream` passes deterministically on all platforms

**Additional context**

Failure output:

```
expected: "d41d8cd98f00b204e9800998ecf8427e"  // MD5("")
actual:   "9a0364b9e99bb480dd25e1f0284c8555"  // MD5("content")
```

`TestStreamLargeFile` (line 289) has the same assertion. It's more stable there because the 2 MiB file blocks the goroutine on the 1 MiB ringbuffer, but the race is the same.

Suggested fix: the comment on line 265/288 says "make sure that no data is being read in order to save on memory". That assertion doesn't actually check that. It checks goroutine scheduling. Drop it. The `io.ReadAll` assertion further down (line 273) still verifies the MD5 end-to-end.

A proper "no buffering on return" guarantee would require `Stream()` to block its internal goroutine until the caller reads.

**Estimation of size**: small

**Estimation of priority**: medium

jhagberg:
This PR closes #683.

**Description**
`TestEncryptTestSuite/TestStream` and `TestStreamLargeFile` were flaky on windows-latest. Both asserted `fs.UnencryptedMD5` equalled MD5("") right after `Stream()` returned. The problem: `Stream` launches a goroutine that updates the hash via `TeeReader` while `io.Copy` runs, so the assertion was really testing goroutine scheduling. On Windows the 7-byte copy sometimes finished before the assertion executed.

The fix drops the two pre-read assertions (plus the `md5.New()` in `TestStreamLargeFile` that went orphan). The `io.ReadAll` assertion further down still verifies the MD5 end-to-end once the stream is drained.

Tracing the root cause surfaced a related issue: `FileStream` exposes four mutable `hash.Hash` fields written by the background goroutine. The only production caller (`upload/upload.go`) reads them after `uploader.Upload` returns, which fully drains `fs.Reader`, and the ringbuffer's mutex gives happens-before. So nothing races in production today. The invariant wasn't documented though, so I added a doc comment on `FileStream` to keep future callers from repeating the mistake the test made.

**How to test**
- \`go test -count=20 -race ./encrypt/\` — 80/80 pass (was flaky on windows-latest)
- \`go test -race ./...\` — 456/456 pass
- \`golangci-lint run ./...\` — clean

### DIFF
--- a/encrypt/encrypt.go
+++ b/encrypt/encrypt.go
@@ -615,6 +615,11 @@ func Stream(file *os.File, pubKeyList [][32]byte) (FileStream, error) {
 	return fs, nil
 }
 
+// FileStream is the output of Stream. The four hash fields are written
+// concurrently by a background goroutine as it copies the source file
+// through crypt4GHWriter. hash.Hash is not safe for concurrent use, so
+// callers must fully drain Reader (read until io.EOF) before calling
+// Sum on any of the hashes.
 type FileStream struct {
 	EncryptedMD5      hash.Hash
 	EncryptedSha256   hash.Hash
--- a/encrypt/encrypt_test.go
+++ b/encrypt/encrypt_test.go
@@ -262,8 +262,6 @@ func (s *EncryptTestSuite) TestStream() {
 
 	fs, err := Stream(file, [][32]byte{s.pubKeyData})
 	assert.NoError(s.T(), err, "failed to create encryption stream")
-	// make sure that no data is being read in order to save on memory
-	assert.Equal(s.T(), hex.EncodeToString(md5hash.Sum(nil)), hex.EncodeToString(fs.UnencryptedMD5.Sum(nil)))
 
 	enc, err := io.ReadAll(fs.Reader)
 	assert.NoError(s.T(), err, "failed to read from encryption stream")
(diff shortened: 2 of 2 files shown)


kostas-kou (approved):

nanjiangshu (approved):
Looks good!

KarlG-nbis (approved):

Issue #677, opened 2026-04-23
feat(download): v2 file download via /files/{fileId} + X-C4GH-Public-Key


jhagberg:
Part of #663. Depends on #676.

**Please describe the feature**
As a data recipient, I want `download --api-version v2` to fetch files by `fileId` through the v2 re-encrypted Crypt4GH path, so my existing CLI commands keep working against the new server.

This is also where `DownloadFile` joins the `Client` interface and the legacy `download.downloadFile` / `getFileIDURL` go away. Deferred from #674 because it requires splitting the existing `downloadFile` (HTTP fetch + progress-bar write) into two pieces.

**Acceptance criteria**
- [ ] `Client` interface: add `DownloadFile(ctx, DownloadRequest) (io.ReadCloser, int64, error)` and `DownloadRequest` type
- [ ] `V1Client.DownloadFile`: absorbs the resolution logic from `download.getFileIDURL` (substring match kept) and the HTTP-fetch half of `download.downloadFile`. Returns the raw body for the caller to stream
- [ ] `V2Client.DownloadFile`:
  - Resolves `UserArg` via `ListFiles({ExactPath: ...})` for paths, list-and-match for bare IDs
  - Follows `FileInfo.DownloadURL` (server-provided; client does not construct `/files/{id}`)
  - Sends `X-C4GH-Public-Key` (base64 of the raw 32-byte key preferred)
  - 403 surfaces as ambiguous "dataset/file does not exist or access denied"
- [ ] `apiclient.File` gains a `DownloadURL` field (empty on v1)
- [ ] `download.writeBodyToDisk(body, totalSize, destPath)` extracted from `downloadFile`. Progress bar and `.part` handling unchanged
- [ ] `datasetCase` / `recursiveCase` / `fileCase` call `client.DownloadFile` + `writeBodyToDisk`
- [ ] Deleted (all callers migrated): `download.GetDatasets`, `download.GetFilesInfo`, `download.File` alias, `download.getFileIDURL`, `download.downloadFile`
- [ ] `download --api-version v2` without `--pubkey` errors client-side with a clear message
- [ ] End-to-end integration test: download the seeded file via v2, check body size matches `Content-Length`
- [ ] Tests verifying the changes are added

**Additional context**
UX change: `--pubkey` is effectively required on v2. v1 currently accepts downloads without it; v2 will not. Document this in the PR description.

**Estimation of size**: big

**Estimation of priority**: high

Issue #658, opened 2026-03-04
Replace deprecated calls from aws-sdk-go-v2


pahatz:
**Describe the bug**
We still use some deprecated funcs, see here:
https://github.com/NBISweden/sda-cli/actions/runs/22570648330/job/65377326635?pr=657
```
  Error: list/list_test.go:89:14: SA1019: manager.NewUploader is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
  	uploader := manager.NewUploader(s3Client)
  	            ^
  Error: list/list_test.go:93:15: SA1019: uploader.Upload is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
  	if _, err := uploader.Upload(context.Background(), &s3.PutObjectInput{
  	             ^
  Error: upload/upload.go:133:14: SA1019: manager.NewUploader is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
  	uploader := manager.NewUploader(s3Client)
  	            ^
  Error: upload/upload.go:224:18: SA1019: uploader.Upload is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
  		result, err := uploader.Upload(ctx, &s3.PutObjectInput{
  		               ^
  Error: upload/upload.go:229:14: SA1019: manager.Uploader is deprecated: superceded by feature/s3/transfermanager. See https://github.com/aws/aws-sdk-go-v2/discussions/3306 (staticcheck)
  		}, func(u *manager.Uploader) {
  		           ^
  5 issues:
  * staticcheck: 5
  
  Error: issues found
``` 
**Steps to reproduce**
Run linter as above
**Expected behavior**
- [ ] clean tests
- [ ] Tests verifying the fix are added

**Additional context**

**Estimation of size**: small/medium/big

**Estimation of priority**: low/medium/high

PR #685, opened 2026-04-24
feat(download): v2 full list — pagination, filters, DatasetInfo

### LINKED ISSUE #675: feat(download): v2 API CI harness + minimal /datasets listing

### LINKED ISSUE #676: feat(download): v2 full list support — pagination, filters, dataset metadata
Part of #663. Depends on #675.

As a data recipient, I want `list` against v2 to paginate results, filter files server-side, and show dataset metadata. Finding a file in a large dataset should be one call, not N+1.

**Acceptance criteria**
- [ ] `apiclient/paginate.go`: generic `paginate[T]` helper. Treats `nil` and empty-string `nextPageToken` both as end-of-stream. Stops on fetch error or cancelled context
- [ ] `V2Client.ListDatasets` now paginates
- [ ] `V2Client.ListFiles(ctx, datasetID, opts)` supports `opts.ExactPath` (`?filePath=<path>`) and `opts.PathPrefix` (`?pathPrefix=<prefix>`). Setting both is a client-side error (friendlier than the server's 400)
- [ ] `V2Client.DatasetInfo(ctx, datasetID)` calls `GET /datasets/{datasetId}`
- [ ] `list --datasets --api-version v2` uses `DatasetInfo` for count and size. No more N+1 file scan
- [ ] `list --dataset <id> <prefix> --api-version v2` pushes the prefix to the server as `PathPrefix`
- [ ] Integration tests: pagination, exact-path filter, path-prefix filter, DatasetInfo

**Additional context**
The v2 pagination params are `pageSize` and `pageToken` (camelCase, per swagger — the prose in `download.md` uses snake_case and is wrong). `nextPageToken` is nullable; represent it as `*string`.

jhagberg:
Closes #676. Stacks on #681. Part of umbrella issue #663.

**Description**
Fills out the v2 client: generic `paginate[T]` helper, paginated `ListDatasets`, `ListFiles` with optional `ExactPath` / `PathPrefix` filters (mutually exclusive), and `DatasetInfo`. Wires the `list` commands to the new endpoints for per-dataset file count + size (replaces v1's N+1 ListFiles loop). Rejects `download --api-version v2` cleanly until #686 wires the real v2 download path so users don't silently half-succeed on `ListFiles` and then fail on the legacy `/s3` transfer.

`V2Client.getJSON` caps the non-2xx error-body read at 201 bytes via `io.LimitReader` — cheap DoS guard until #687 replaces the error path with typed RFC 9457 Problem Details.

Carries the field-rename + `int64` widening from the #679 review-amend through the v2 layer: `int(f.DecryptedSize)` in `v2File.toFile()` drops to plain `f.DecryptedSize`, `formatFileSizeOutput` receives `DatasetInfo.Size` directly without the temporary `int` cast, and `paginate` `ListDatasets` keeps the `url.JoinPath` form from #681's review-amend.

**How to test**
- `go build ./...`, `go vet ./...`, `go test ./...` — 186 pass.
- Integration: `go test -tags=integration ./download/... -run 'TestV2_'` against the live dev stack — 5 smoke tests pass.
- E2E with a rebuilt `sda-cli`:
  - `list --datasets --api-version v2` prints per-dataset count + size from `DatasetInfo`.
  - `list --dataset X --api-version v2` prints the file list.
  - `list --dataset X --api-version v2 <prefix>` filters server-side via `pathPrefix`.
  - `download --api-version v2 ...` errors with `v2 download is not yet implemented (see #677)`.

## Test plan
- [x] `go test ./...`: 186 pass
- [x] `go vet ./...`: clean
- [x] `golangci-lint run`: clean (5 inherited `SA1019` hits tracked in #658)
- [x] `gofmt -l .`: clean
- [x] Integration: `TestV2_ListDatasets_Smoke`, `TestV2_ListFiles_Smoke`, `TestV2_ListFiles_ExactPath_Smoke`, `TestV2_ListFiles_PathPrefix_NoMatch`, `TestV2_DatasetInfo_Smoke` all pass
- [x] Manual: per-dataset enrichment, exact-path + prefix filters, transport-error wrap, unsupported version, download-v2 guard

### DIFF
--- a/downloadclient/v2.go
+++ b/downloadclient/v2.go
@@ -46,7 +46,7 @@ func (c *V2Client) ListDatasets(ctx context.Context) ([]string, error) {
 	return paginate(ctx, func(ctx context.Context, pageToken *string) ([]string, *string, error) {
 		u := endpoint
 		if pageToken != nil {
-			u += "?pageToken=" + url.QueryEscape(*pageToken)
+			u += "?" + url.Values{"pageToken": {*pageToken}}.Encode()
 		}
 		body, err := c.getJSON(ctx, u)
 		if err != nil {
@@ -63,9 +63,48 @@ func (c *V2Client) ListDatasets(ctx context.Context) ([]string, error) {
 	})
 }
 
-// ListFiles implements Client. Not implemented until #676.
-func (c *V2Client) ListFiles(_ context.Context, _ string, _ ListFilesOptions) ([]File, error) {
-	return nil, errors.New("V2Client.ListFiles not implemented until #676")
+// ListFiles implements Client. Walks all pages of GET /datasets/{id}/files,
+// optionally applying the v2 server-side filters (exact filePath, or recursive
+// pathPrefix). The two filters are mutually exclusive per v2's contract;
+// we reject that combo client-side for a friendlier message than the
+// server's 400.
+func (c *V2Client) ListFiles(ctx context.Context, datasetID string, opts ListFilesOptions) ([]File, error) {
+	if opts.ExactPath != "" && opts.PathPrefix != "" {
+		return nil, errors.New("ListFilesOptions.ExactPath and .PathPrefix are mutually exclusive")
+	}
+
+	return paginate(ctx, func(ctx context.Context, pageToken *string) ([]File, *string, error) {
+		u := c.cfg.BaseURL + "/datasets/" + url.PathEscape(datasetID) + "/files"
+		q := url.Values{}
+		if opts.ExactPath != "" {
+			q.Set("filePath", opts.ExactPath)
+		}
+		if opts.PathPrefix != "" {
+			q.Set("pathPrefix", opts.PathPrefix)
+		}
+		if pageToken != nil {
+			q.Set("pageToken", *pageToken)
+		}
+		if enc := q.Encode(); enc != "" {
+			u += "?" + enc
+		}
+		body, err := c.getJSON(ctx, u)
+		if err != nil {
+			return nil, nil, err
+		}
+		defer body.Close() //nolint:errcheck
+
+		var resp fileListResponse
+		if err := json.NewDecoder(body).Decode(&resp); err != nil {
+			return nil, nil, fmt.Errorf("failed to decode /datasets/%s/files response: %w", datasetID, err)
+		}
+		out := make([]File, len(resp.Files))
+		for i, f := range resp.Files {
+			out[i] = f.toFile()
+		}
+
+		return out, resp.NextPageToken, nil
+	})
 }
 
 // DatasetInfo implements Client. Not implemented until #676.
--- a/downloadclient/v2.go
+++ b/downloadclient/v2.go
@@ -43,6 +43,7 @@ func (c *V2Client) ListDatasets(ctx context.Context) ([]string, error) {
 	if err != nil {
 		return nil, fmt.Errorf("invalid base URL: %w", err)
 	}
+
 	return paginate(ctx, func(ctx context.Context, pageToken *string) ([]string, *string, error) {
 		u := endpoint
 		if pageToken != nil {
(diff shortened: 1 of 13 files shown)


downloadclient/v2.go
```diff
@@ -29,39 +29,98 @@ func NewV2Client(cfg Config) *V2Client {
...
+// pathPrefix). The two filters are mutually exclusive per v2's contract;
+// we reject that combo client-side for a friendlier message than the
+// server's 400.
+func (c *V2Client) ListFiles(ctx context.Context, datasetID string, opts ListFilesOptions) ([]File, error) {
+	if opts.ExactPath != "" && opts.PathPrefix != "" {
+		return nil, errors.New("ListFilesOptions.ExactPath and .PathPrefix are mutually exclusive")
+	}
+
+	return paginate(ctx, func(ctx context.Context, pageToken *string) ([]File, *string, error) {
+		u := c.cfg.BaseURL + "/datasets/" + url.PathEscape(datasetID) + "/files"
```
nanjiangshu: Would be nice to use a library for path join here: 

e.g. 

```golang
u, err := url.JoinPath(c.cfg.BaseURL, "datasets", datasetID, "files")
```

The same as in `DatasetInfo`

nanjiangshu (approved):
Looks good to me. I just have a tiny comment about url concatenation but this does not hinder an approval.

downloadclient/v2.go
jhagberg: Good catch. Small wrinkle: `DatasetInfo` (line 108) also uses string concat with `url.PathEscape`, so it is not on `url.JoinPath` either. Only `ListDatasets` is.

Switching both `ListFiles` and `DatasetInfo` to `url.JoinPath` drops the `url.PathEscape(datasetID)` wrap, because `JoinPath` does not percent-encode segments. SDA dataset IDs are EGAD-prefixed alphanumerics in practice so the escape is dead code, and a hostile `--dataset-id` would only affect the user's own session. Will switch both to `JoinPath` if that's fine, otherwise happy to keep the existing concat+`PathEscape` style.

jbygdell: As per https://github.com/neicnordic/sensitive-data-archive/blob/7190f716c9b4e956b2be8fee8c1dfe31b871b2e6/sda/schemas/isolated/dataset-mapping.json#L26 the dataset ID can be any string, even a DOI with the format `NN.NNNN/NNN` so escaping any special characters in the dataset identifier needs to be an option.

jhagberg: Aha good point @jbygdell, the schema does allow DOI-format. `url.PathEscape` stays as a real necessity, not dead code, and I withdraw the "EGAD-only in practice" claim :)

Verified with a quick stdlib repro: a plain `url.JoinPath(c.cfg.BaseURL, "datasets", datasetID, "files")` does turn `10.1234/567` into two path segments and 404s. But `url.JoinPath(c.cfg.BaseURL, "datasets", url.PathEscape(datasetID), "files")` preserves the `%2F` via `RawPath` and produces the right URL, so the `ListDatasets` style is reachable, just a bit more verbose than the current concat. @nanjiangshu @jbygdell let me know which you prefer.

jbygdell: `url.JoinPath(c.cfg.BaseURL, "datasets", url.PathEscape(datasetID), "files")` would be my choice since it will generate a clean URL path.

jhagberg: @nanjiangshu ?

nanjiangshu: Yes, Jockes suggestions looks good to me.

jhagberg: Done :) Switched both `ListFiles` and `DatasetInfo` to `url.JoinPath(c.cfg.BaseURL, "datasets", url.PathEscape(datasetID), ...)` to match the `ListDatasets` pattern, keeping `PathEscape` per @jbygdell's schema note about DOI-format IDs. Folded into pr3's review-amend (commit `8de92bc`) and cascaded the rebase onto #686 and #687.

jbygdell (approved):

PR #686, opened 2026-04-24
feat(download): v2 file download via /files/{fileId}

### LINKED ISSUE #677: feat(download): v2 file download via /files/{fileId} + X-C4GH-Public-Key

jhagberg:
Closes #677. Stacks on #685.

**Description**
Wires v2 file download through `GET /files/{fileId}` plus the server-provided `downloadUrl` plus the `X-C4GH-Public-Key` header, retires the legacy `/s3`-transfer helpers, and shrinks `download/download.go` to a `downloadOne` → `writeBodyToDisk` flow sitting on `apiclient.Client.DownloadFile`. Removes the "v2 download is not yet implemented" guard that #685 put in.

**Key design points**

- `Client.DownloadFile(ctx, DownloadRequest) (DownloadResult, error)` is the new abstraction. `DownloadResult` carries the canonical `File` (authoritative filename — UserArg can be a bare fileId), the response body, and the server's `Content-Length`.
- `V2Client.DownloadFile` resolves UserArg via `ExactPath` (paths) or `ListFiles` scan (bare ids), then hits the server-provided `downloadUrl` with `X-C4GH-Public-Key`. URLs go through `url.Parse` + `ResolveReference` so both relative and absolute forms work.
- **Bearer-token leakage guard:** when the resolved download host differs from `BaseURL`, the `Authorization` header is dropped. A pre-signed cross-origin URL is self-authenticating; leaking the bearer to a third-party host is the kind of thing you read about in CVE writeups.
- 403 from either resolve or GET flattens to `dataset/file does not exist or access denied: <arg>` — existence-leakage contract preserved from v1.
- `downloadOne` uses `result.File.FilePath` (anonymized) for the on-disk path. Downloading by fileId now produces a meaningfully-named file instead of `outdir/<fileId>`. A containment check rejects any `outputPath` escaping `outDir`, so a server-provided path with `../` can't write outside the configured output directory.
- Retires `downloadFile`, `getFileIDURL`, `GetDatasets`, `GetFilesInfo`, `getBody`, `setupCookieJar`, the `download.File` alias, and the `cookieJar`/`cookiePath`/`appVersion` package vars. The `apiclient.WithV1CookieJar` option also retires here in the same refactor.
- Carries the field rename (`Version` → `ClientVersion`) and `int64` widening from #679's review-amend into the new V1/V2 DownloadFile sites, and mirrors #681's `SDA-Client-Version` + `User-Agent` header pair on v2.

**Out of scope (tracked separately)**

- N+1 re-resolve on `datasetCase` / `recursiveCase` (separate follow-up).
- RFC 9457 Problem Details — #687.

**How to test**

Local:
- `go build ./...`, `go vet ./...`, `gofmt -l .`: clean.
- `go test ./...`: 176 pass.
- `golangci-lint run --timeout 5m`: 0 issues.

Integration (build tag `integration`, against the live v2 dev stack):
- `TestV2_ListDatasets_Smoke`, `TestV2_ListFiles_Smoke`, `TestV2_ListFiles_ExactPath_Smoke`, `TestV2_ListFiles_PathPrefix_NoMatch`, `TestV2_DatasetInfo_Smoke`, `TestV2_DownloadFile_EndToEnd`: all pass.

Manual, rebuilt `sda-cli` against the live dev stack:
- `download --api-version v2 --dataset-id EGAD00000000001 --pubkey <pem> test-file.c4gh` writes the file (~1.2 kB c4gh bytes).
- `download --api-version v2 ... EGAF00000000001` (by fileId) writes `test-file.c4gh` — the canonical name, not the bare fileId.
- `download --api-version v2` without `--pubkey` errors with `v2 downloads require --pubkey`.
- `download --api-version v2 --dataset-id DOES_NOT_EXIST ...` errors with `dataset/file does not exist or access denied: <arg>` (403 flattened).

## Test plan
- [x] `go test ./...`: 176 pass
- [x] `go vet ./...`: clean
- [x] `gofmt -l .`: clean
- [x] Integration: end-to-end v2 download passes
- [x] Manual: by-path, by-fileId, missing pubkey, 403 flatten

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -3,15 +3,11 @@ package download
 import (
 	"bufio"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
-	"slices"
 	"strings"
 	"time"
 
@@ -21,8 +17,6 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/vbauerster/mpb/v8"
 	"github.com/vbauerster/mpb/v8/decor"
-	"go.nhat.io/cookiejar"
-	"golang.org/x/net/publicsuffix"
 )
 
 var datasetID string
@@ -40,11 +34,11 @@ var apiVersionFlag string
 var downloadCmd = &cobra.Command{
 	Use:   "download [flags] [filepath(s) | fileid(s)]",
 	Short: "Download files from SDA",
-	Long: `Download files from the Sensitive Data Archive (SDA) using APIs at the specified URL. 
+	Long: `Download files from the Sensitive Data Archive (SDA) using APIs at the specified URL.
 	The user must have the necessary access rights (visas) to the datasets being downloaded
 	Important:
 		Provide exactly one of the following options to specify files to download:
-			- [filepath(s) or fileid(s)] 
+			- [filepath(s) or fileid(s)]
 			- --dataset
 			- --recursive <dirpath>
 			- --from-file <list-filepath>`,
@@ -73,45 +67,12 @@ func init() {
 	downloadCmd.Flags().StringVar(&apiVersionFlag, "api-version", "v1", "SDA download API version to use (v1 or v2)")
 }
 
-var cookieJar *cookiejar.PersistentJar
-var cookiePath string
-var appVersion string
-
-// File is the file metadata type. Canonical definition lives in apiclient.
-// Alias preserves source compat for existing callers (list/, tests). Removed
-// in #677 when callers reference apiclient.File directly.
-type File = apiclient.File
-
 // Download function downloads files from the SDA by using the
 // download's service APIs
 func Download(args []string, configPath, version string) error {
-	appVersion = version
-
 	if datasetID == "" || URL == "" || configPath == "" {
 		return errors.New("missing required arguments, dataset-id, config and url are required")
 	}
-
-	// Fail fast on an unsupported --api-version before we touch the
-	// filesystem via setupCookieJar. Cheap check; avoids creating
-	// ${UserCacheDir}/sda-cli/ when the command is about to error out.
-	if err := apiclient.ValidateVersion(apiVersionFlag); err != nil {
-		return err
-	}
-
-	// v2 list commands work as of #676, but the actual download still goes
-	// through the legacy /s3 transfer path. Without this guard, a successful
-	// v2 ListFiles would be followed by a silent 404 on /s3/{fileID} once
-	// the real v2 dev stack is targeted. Reject cleanly until #677 wires
-	// v2 download via /files/{fileId} + X-C4GH-Public-Key.
-	if apiVersionFlag == "v2" {
-		return errors.New("v2 download is not yet implemented (see #677)")
-	}
-
-	u, err := url.Parse(URL)
-	if err != nil || u.Scheme == "" {
-		return errors.New("invalid base URL")
-	}
-	setupCookieJar(u)
 
 	// Check if both --ignore-existing and --overwrite-existing are set
 	if ignoreExisting && overwriteExisting {
@@ -155,16 +116,11 @@ func Download(args []string, configPath, version string) error {
 		return err
 	}
 
-	// Share the cookie jar that downloadFile uses so V1Client's metadata
-	// calls and the legacy /s3 transfer path see the same in-memory
-	// cookie state. Two independent AutoSync:ed jars on the same on-disk
-	// file would race and clobber each other (fixed here; removed in
-	// #677 when downloadFile also moves onto apiclient.Client).
 	client, err := apiclient.New(apiclient.Config{
 		BaseURL: URL,
 		Token:   config.AccessToken,
 		Version: version,
-	}, apiVersionFlag, apiclient.WithV1CookieJar(cookieJar))
+	}, apiVersionFlag)
 	if err != nil {
 		return err
 	}
@@ -174,22 +130,22 @@ func Download(args []string, configPath, version string) error {
 
 	switch {
 	case datasetDownload:
-		err = datasetCase(ctx, client, config.AccessToken)
+		err = datasetCase(ctx, client)
 		if err != nil {
 			return err
 		}
 	case recursiveDownload:
-		err = recursiveCase(ctx, client, args, config.AccessToken)
+		err = recursiveCase(ctx, client, args)
 		if err != nil {
 			return err
 		}
 	case fromFile:
-		err = fileCase(ctx, client, args, config.AccessToken, true)
+		err = fileCase(ctx, client, args, true)
 		if err != nil {
 			return err
 		}
 	default:
-		err = fileCase(ctx, client, args, config.AccessToken, false)
+		err = fileCase(ctx, client, args, false)
 		if err != nil {
 			return err
 		}
@@ -198,7 +154,7 @@ func Download(args []string, configPath, version string) error {
 	return nil
 }
 
-func datasetCase(ctx context.Context, client apiclient.Client, token string) error {
+func datasetCase(ctx context.Context, client apiclient.Client) error {
 	fmt.Println("Downloading all files in the dataset")
 	files, err := client.ListFiles(ctx, datasetID, apiclient.ListFilesOptions{})
 	if err != nil {
@@ -206,11 +162,7 @@ func datasetCase(ctx context.Context, client apiclient.Client, token string) err
 	}
 
 	for _, file := range files {
-		fileName := helpers.AnonymizeFilepath(file.FilePath)
-		fileURL := URL + "/s3/" + datasetID + "/" + fileName
-
-		err = downloadFile(fileURL, token, pubKeyBase64, file.FilePath)
-		if err != nil {
+		if err := downloadOne(ctx, client, file.FilePath); err != nil {
 			return err
 		}
 	}
@@ -218,12 +170,8 @@ func datasetCase(ctx context.Context, client apiclient.Client, token string) err
 	return nil
 }
 
-func recursiveCase(ctx context.Context, client apiclient.Client, args []string, token string) error {
+func recursiveCase(ctx context.Context, client apiclient.Client, args []string) error {
 	fmt.Println("Downloading content of the path(s)")
-	files, err := client.ListFiles(ctx, datasetID, apiclient.ListFilesOptions{})
-	if err != nil {
-		return fmt.Errorf("failed to get files, reason: %v", err)
-	}
 
 	var dirPaths []string
 	for _, path := range args {
@@ -232,25 +180,44 @@ func recursiveCase(ctx context.Context, client apiclient.Client, args []string, 
 		}
 		dirPaths = append(dirPaths, path)
 	}
+
+	// v1 has no server-side prefix filter, so fetch the full list once and
+	// filter client-side. v2 has a pathPrefix filter; apply it per dirPath.
+	var allFiles []apiclient.File
+	if apiVersionFlag != "v2" {
+		files, err := client.ListFiles(ctx, datasetID, apiclient.ListFilesOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get files, reason: %v", err)
+		}
+		allFiles = files
+	}
+
 	var missingPaths []string
-	// Loop over all the files of the dataset and
-	// check if the provided path is part of their filepath.
-	// If it is then download the file
 	for _, dirPath := range dirPaths {
-		pathExists := false
-		for _, file := range files {
-			if strings.HasPrefix(file.FilePath, dirPath) {
-				pathExists = true
-				fileName := helpers.AnonymizeFilepath(file.FilePath)
-				fileURL := URL + "/s3/" + datasetID + "/" + fileName
-				err = downloadFile(fileURL, token, pubKeyBase64, file.FilePath)
-				if err != nil {
-					return err
+		var matched []apiclient.File
+		if apiVersionFlag == "v2" {
+			files, err := client.ListFiles(ctx, datasetID, apiclient.ListFilesOptions{PathPrefix: dirPath})
+			if err != nil {
+				return fmt.Errorf("failed to get files, reason: %v", err)
+			}
+			matched = files
+		} else {
+			for _, f := range allFiles {
+				if strings.HasPrefix(f.FilePath, dirPath) {
+					matched = append(matched, f)
 				}
 			}
 		}
-		if !pathExists {
+
+		if len(matched) == 0 {
 			missingPaths = append(missingPaths, dirPath)
+
+			continue
+		}
+		for _, file := range matched {
+			if err := downloadOne(ctx, client, file.FilePath); err != nil {
+				return err
+			}
 		}
 	}
 	if len(missingPaths) == len(dirPaths) {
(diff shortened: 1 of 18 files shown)


downloadclient/types.go
```diff
@@ -13,6 +13,9 @@ type File struct {
 	DecryptedFileSize         int64  `json:"decryptedFileSize"`
 	DecryptedFileChecksum     string `json:"decryptedFileChecksum"`
 	DecryptedFileChecksumType string `json:"decryptedFileChecksumType"`
+	// DownloadURL is v2-only (server-provided relative URL).
+	// Empty on v1 File values.
+	DownloadURL string `json:"-"`
```
KarlG-nbis: Why does downloadv2 need to expose this? If its only for use within the downloadclient package, it shouldnt be exposed outside(ie not start with capital character)

downloadclient/v2.go
```diff
@@ -113,6 +112,125 @@ func (c *V2Client) ListFiles(ctx context.Context, datasetID string, opts ListFil
...
+
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodGet, resolved.String(), nil)
+	if err != nil {
+		return DownloadResult{}, err
+	}
+	// Authorization is only safe on the BaseURL host. If the server
+	// redirects to a different origin (e.g. a pre-signed S3 URL) the
+	// presigned signature is self-authenticating and leaking the bearer
+	// token to that host is both unnecessary and risky.
+	if resolved.Host == base.Host {
```
KarlG-nbis: Why is this the case? So if server redirects to other origin we dont set `Authorization`? In which scenarios do we expect that? Wouldnt it to better to just error in that case?

downloadclient/types.go
jhagberg: In v1 the client constructs the download URL itself by convention (`/s3/{dataset}/{filePath}`, see `v1.go:185`). The v2 API changed this so the server returns a `downloadUrl` per file in the listing response (see the swagger `FileInfo` schema and our wire type in `v2_types.go:16`). The client then just follows that URL instead of knowing the URL pattern.

So `DownloadURL` on `File` is how v2 carries the server-provided URL from `ListFiles` through `resolveFile` into `DownloadFile` (`v2.go:145-160`). V1 never populates it, which is why it is tagged `json:"-"` and the comment says "Empty on v1 File values."

An alternative would be a separate internal struct just for the resolve step, but it felt like unnecessary indirection since `File` already carries all the metadata and `resolveFile` needs to return both. Happy to reconsider if you think it should be split out though?

downloadclient/v2.go
```diff
@@ -113,6 +112,125 @@ func (c *V2Client) ListFiles(ctx context.Context, datasetID string, opts ListFil
...
+			return File{}, err
+		}
+		if len(files) == 0 {
+			return File{}, nil
+		}
+
+		return files[0], nil
+	}
+	// Bare id: list + match. v2 has no exact-id filter.
+	files, err := c.ListFiles(ctx, datasetID, ListFilesOptions{})
```
KarlG-nbis, line 215: Wouldnt it be "better" to just called `HEAD c.cfg.BaseUrl//files/$userArg` in this case and to check if file exists (and user has access), then to list through all files in a dataset, and then just construct the File.DownloadUrl based on that?

downloadclient/types.go
```diff
@@ -13,6 +13,9 @@ type File struct {
 	DecryptedFileSize         int64  `json:"decryptedFileSize"`
 	DecryptedFileChecksum     string `json:"decryptedFileChecksum"`
 	DecryptedFileChecksumType string `json:"decryptedFileChecksumType"`
```
KarlG-nbis: Is `DecryptedFileChecksum`, and `DecryptedFileChecksumType` needed, can just see them being used from tests, or do we marshal File structs somewhere and give that to the caller?

downloadclient/types.go
jhagberg: Sorry, my first reply was for the original version of your comment! You are right, `DownloadURL` is only used within the `downloadclient` package so it should not be exported. Fixed it now, renamed to `downloadURL` in the latest push.

downloadclient/v2.go
jhagberg: The v2 server can return an absolute `downloadUrl` pointing to a pre-signed storage URL (e.g. S3 with `X-Amz-Signature` in the query string). In that case the pre-signed URL is already self-authenticating, and sending our bearer token to a third-party host would be a credential leak. The host check follows the same pattern as Go's own `net/http` client which strips Authorization headers on cross-origin redirects. Erroring would break a valid server implementation since the v2 swagger allows both relative and absolute download URLs.

downloadclient/v2.go
jhagberg: You are right that listing all files to find one by ID is not great. But we cannot construct the download URL client-side since the v2 contract says downloadUrl is server-provided (could be a relative path or a pre-signed link).

When the caller knows the file path we already use the exact ?filePath filter which returns just that one file. The list-all fallback only fires for bare fileIds, and that is the same approach v1 uses (v1MatchFile in v1.go:228). I am not sure but does not most users download by path since they get it from sda-cli list output?

The v2 API does not have a single-file-by-id lookup endpoint. We could file an issue for GET /datasets/{id}/files/{fileId} against sensitive-data-archive if you think it is worth adding, or we can just keep the v1 behavior for now. What do you think?

downloadclient/types.go
jhagberg: Good catch. Nothing in the CLI reads these fields today, `list.go` only uses `FileID`, `DecryptedFileSize`, and `FilePath`. They are on the struct because `File` mirrors the v1 JSON wire shape and v1 deserializes them from the server response. In v2 `toFile()` maps from the checksums array to keep the struct populated consistently.

I kept them because they seem useful for post-download checksum verification later, but you are right that right now they are not consumed outside of tests. Do you think we should drop them, or keep them for when we add verification?

KarlG-nbis: I think it makes sense to remove if not used currently, and we can add them back if to be used later.

downloadclient/v2.go
KarlG-nbis: Sure following what the contract says keeping the current behaviour makes sense, but in practice with the current download-v2 implementation wouldnt we be able to construct the download URL?

But yeah makes sense to follow what the contract says

downloadclient/v2.go
KarlG-nbis: Yeah sure that makes sense if we follow the contract ([but it doesnt mention anything about the pre-resolved url to be self-authenticating](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/download/swagger_v2.yml#L1016)), but perhaps Im missing something. But in what scenarios would we expect such behaviour?

downloadclient/v2.go
```diff
@@ -113,6 +112,125 @@ func (c *V2Client) ListFiles(ctx context.Context, datasetID string, opts ListFil
...
+	// presigned signature is self-authenticating and leaking the bearer
+	// token to that host is both unnecessary and risky.
+	if resolved.Host == base.Host {
+		httpReq.Header.Set("Authorization", "Bearer "+c.cfg.Token)
+	}
+	httpReq.Header.Set("X-C4GH-Public-Key", req.PublicKeyBase64)
+	if c.cfg.ClientVersion != "" {
+		httpReq.Header.Set("User-Agent", "sda-cli/"+c.cfg.ClientVersion)
+		httpReq.Header.Set("SDA-Client-Version", c.cfg.ClientVersion)
+	}
```
KarlG-nbis, line 172: it makes sense to add these for debugging, but looking into the download-v2 implementation can not see that we check the `SDA-Client-Version` value similar as what the download v1 implementation did, was that intentionally skipped?

downloadclient/v2.go
KarlG-nbis: Also you mentioned that the `downloadUrl` could be a pre-signed link but this does not appear to be mentioned in the [swagger](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/download/swagger_v2.yml#L822), per the swagger it should only be a relative path

jhagberg: Yes you are right.  The pre-signed link scenario is a speculating thought Idea I have had, that is not in the spec.

jhagberg: So yes, with the current spec you could construct /files/{fileId} directly. The reason I still use the server-provided value is just to avoid hardcoding the URL pattern in the client, but I agree the argument is weaker than I made it sound.

downloadclient/v2.go
jhagberg: You are right, the swagger says downloadUrl is a relative path so the resolved URL will always have the same host as BaseURL. The host check is dead code with the current spec. I will remove it and simplify to always set Authorization.

downloadclient/v2.go
KarlG-nbis: I think its worth keeping as is, but possibly clarifying current behaviour and future thinking in a comment, etc to reduce of risk confusion of how the server behaves, etc

downloadclient/v2.go
jhagberg: Good question. The v1 download server (`sda-download/`) has a `ClientVersionMiddleware` that checks `SDA-Client-Version` against a configured minimum and returns 412 if missing or too old. The v2 download server (`sda/cmd/download/`) does not have that middleware, so the header is not enforced server-side on v2.

It was not an intentional decision to skip it, it just was not part of the v2 server implementation yet. We still send it from the client so it shows up in server logs for debugging and tracing. Do you think we should add a version check middleware to the v2 server as well?

downloadclient/types.go
jhagberg: Will remove them

downloadclient/v2.go
jhagberg: Yes I will add something. What do you think about filing an issue?

downloadclient/v2.go
KarlG-nbis: Also not fully aligned with comment

```
ClientVersion string // sda-cli version (e.g. v1.2.3); SDA-Client-Version on v1, User-Agent "sda-cli/<version>" on v2
```

as `SDA-Client-Version` is also sent on v2

KarlG-nbis: It makes sense to add to download-v2 to support rejecting a client running an older version if we've added breaking changes to ensure client works as expected

jhagberg: Agreed, makes sense to have that safety net on v2 as well. That would be a server-side change in sensitive-data-archive though, not this PR. I can file an issue for it if you want?

jhagberg: Filed neicnordic/sensitive-data-archive#2442 for adding the version check middleware to the v2 server.

downloadclient/v2.go
KarlG-nbis: I currently do not see a use case where we would want to support a pre-signed url that give the holder of the url access to a file, as seems quite tricky to keep such functionality secure to not risk leaking sensitive data. So would suggest to check with @viklund if this is a feature we want

KarlG-nbis (approved):

downloadclient/v2.go
jhagberg: I meant this "The v2 API does not have a single-file-by-id lookup endpoint. We could file an issue for GET /datasets/{id}/files/{fileId} against sensitive-data-archive if you think it is worth adding"?

viklund: Yes, seems like a good API endpoint to have.

downloadclient/types.go
```diff
@@ -2,17 +2,15 @@
...
-	DecryptedFileSize         int64  `json:"decryptedFileSize"`
-	DecryptedFileChecksum     string `json:"decryptedFileChecksum"`
-	DecryptedFileChecksumType string `json:"decryptedFileChecksumType"`
+	FileID            string `json:"fileId"`
+	DisplayFileName   string `json:"displayFileName"`
+	FilePath          string `json:"filePath"`
+	DecryptedFileSize int64  `json:"decryptedFileSize"`
+	// downloadURL is v2-only (server-provided relative URL).
+	// Empty on v1 File values.
+	downloadURL string
```
nanjiangshu, line 13:
```suggestion
	DownloadURL string `json:"downloadUrl,omitempty"`
```
I think this needs to be exported since other wise json.Unmarshal will sliently ignore it.

download/download.go
```diff
@@ -232,25 +180,44 @@ func recursiveCase(ctx context.Context, client downloadclient.Client, args []str
...
 		}
-		if !pathExists {
+
+		if len(matched) == 0 {
 			missingPaths = append(missingPaths, dirPath)
+
+			continue
+		}
+		for _, file := range matched {
+			if err := downloadOne(ctx, client, file.FilePath); err != nil {
```
nanjiangshu, line 218: In this loop, downloadOne will call 
```
target, err := c.resolveFile(ctx, req.DatasetID, req.UserArg)
```

and further call

`c.ListFiles`

This means downloading e.g. 10k files will request the network pigination 10k time. It's a lot of overhead.

Since `datasetCase` and `recursiveCase` already have the resolved File struct, downloadOne should accept a `downloadclient.File` parameter if available, bypassing the redundant path resolution inside DownloadFile.

nanjiangshu (commented):
Looks good in general. I left one major comment regarding the unnecessary network overhead when downloading multiple files of a dataset. Another minor comment is about exporting of downloadURL in the struct.

downloadclient/types.go
jhagberg: Good thought, but `downloadURL` on `File` is not populated via `json.Unmarshal`. The v2 response is decoded into the internal `v2File` wire type (which has an exported `DownloadURL` with the json tag in `v2_types.go`), and then `toFile()` copies it over programmatically. For v1, the server does not return a `downloadUrl` field at all so the field would always be empty even if exported.

Karl asked us to unexport it since it is only used within the `downloadclient` package. Does that make sense or do you see a case I am missing?

download/download.go
jhagberg: You are right, this is a real overhead for large datasets. The `datasetCase` and `recursiveCase` paths already have the resolved files but then `downloadOne` re-resolves each one via `resolveFile`. I think adding a `DownloadFileByMetadata` method (or similar) that accepts an already-resolved `File` is the right fix, but it is a bigger change than what this PR covers. Should I file an issue to track it?

nanjiangshu: Sounds good to address it in a new issue.

nanjiangshu (approved):
Looks good to me now.

downloadclient/v2.go
jhagberg: Filed neicnordic/sensitive-data-archive#2443 for the single-file-by-id endpoint.

download/download.go
jhagberg: Filed #698 for this. @nanjiangshu does the issue description look right to you?

PR #687, opened 2026-04-24
feat(apiclient): RFC 9457 Problem Details on v2 error responses

### LINKED ISSUE #678: feat(download): RFC 9457 Problem Details error parsing on v2 calls
Part of #663. Depends on #677.

As a data recipient on v2, I want server errors shown as readable prose instead of raw JSON or bare status codes. 403 responses should stay ambiguous about whether the file does not exist or I lack access — that's the server's contract.

**Acceptance criteria**
- [ ] `apiclient/errors.go`: `ProblemDetails` type (`type`, `title`, `status`, `detail`, `instance`) and `parseErrorResponse(*http.Response) error`
- [ ] Parser handles: valid full Problem Details, title-only, detail-only, invalid JSON, empty body, wrong content-type. No misleading messages in any case
- [ ] Wired into every v2 call site: `V2Client.getJSON` and `V2Client.DownloadFile` non-2xx paths
- [ ] 403 on resource-by-id endpoints (`/datasets/{id}`, `/files/{id}`, `/files/{id}/content`) always surfaces as "dataset/file does not exist or access denied". The server's `Title` / `Detail` is not echoed if it would distinguish the two cases
- [ ] Integration test: fetch a nonexistent fileId on v2, assert the ambiguous prefix in the error

**Additional context**
Server contract: [RFC 9457](https://datatracker.ietf.org/doc/html/rfc9457) `application/problem+json`. Resource-by-id endpoints collapse "not found" and "access denied" into 403 so the existence of an archived file is not leaked to someone without access.

**Estimation of size**: small

**Estimation of priority**: medium

jhagberg:
Closes #678. Stacks on #686. Fifth of five PRs for #663.

**Description**
Replaces `V2Client`'s ad-hoc `"server returned status N: {body}"` error format with RFC 9457 Problem Details parsing, adds a typed `APIError` with `StatusCode` for `errors.As`-based callers, and drops the substring-match hack in `V2Client.DownloadFile`'s 403-flatten logic. Small PR (+205 / −13), focused on the error surface; success path unchanged.

**Key design points**

- `apiclient.APIError` is a typed error with `StatusCode` (so callers can `errors.As`), an optional `*ProblemDetails` populated when the body parses, and a truncated `Body` for diagnostics. The `msg` field stays unexported so the formatted message always goes through `Error()`.
- `parseErrorResponse` tries Problem Details when Content-Type is `application/problem+json` or `application/json` (case-insensitive via `mime.ParseMediaType`). Falls back to `"server returned status N: <truncated>"` for HTML / plain / empty / malformed. Body read capped at 8 KiB.
- `V2Client.DownloadFile`'s resolve-step 403 flatten moves from `strings.Contains(err.Error(), "status 403")` to `errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusForbidden`. The old substring match silently stopped firing the moment the server started returning Problem Details — the new message contains `"Forbidden (403)"`, never `"status 403"` — so the 403 would have leaked. This PR is what `getJSON`'s `// #678 replaces error wrapping...` comment was waiting for.
- Both 403 flatten sites still produce `"dataset/file does not exist or access denied: <UserArg>"`; the typed `*APIError` message is discarded on that path so the leakage contract can't leak Problem Details detail.

**User-visible CLI message change (worth flagging at review)**

Non-403 v2 server responses change shape. Dev stack returns `application/problem+json`, so:

- Old: `Error: failed to get datasets, reason: server returned status 401: {"detail":"Missing, invalid, or expired bearer token.",...}`
- New: `Error: failed to get datasets, reason: Unauthorized (401): Missing, invalid, or expired bearer token.`

403 flattening is unchanged. Shell scripts or log filters keyed on the literal `"server returned status N"` need updating for the Problem-Details path.

**Out of scope (tracked separately)**

- Required-field validation on `v2File` (deferred Codex finding).
- Clean `"invalid base URL"` shape for empty/malformed `BaseURL` on v2.
- N+1 re-resolve on `datasetCase` / `recursiveCase` (carried from #686).

**How to test**

Local:
- `go build ./...`, `go vet ./...`, `gofmt -l .`: clean.
- `go test ./...`: 191 pass.
- `golangci-lint run --timeout 5m`: 0 issues.

Integration (build tag `integration`, against the live v2 dev stack):
- `TestV2_ListDatasets_Smoke`, `TestV2_ListFiles_Smoke`, `TestV2_ListFiles_ExactPath_Smoke`, `TestV2_ListFiles_PathPrefix_NoMatch`, `TestV2_DatasetInfo_Smoke`, `TestV2_DownloadFile_EndToEnd`, `TestV2_DownloadFile_NotFound403`: 7/7 pass.

Manual against the live dev stack:
- `list --api-version v2 --dataset EGAD_FORBIDDEN` → `"Forbidden (403): access denied"` (typed; list has no leakage flatten).
- `list --api-version v2 --datasets` with fake-signature JWT → `"Unauthorized (401): Missing, invalid, or expired bearer token."`
- `download --api-version v2 --dataset-id EGAD_DOES_NOT_EXIST --pubkey <pem> some-file.c4gh` → `"dataset/file does not exist or access denied: some-file.c4gh"` (403 on list-resolve via the typed `errors.As` path — would have leaked under the old substring hack).
- Same with bare fileId → same flatten message.
- Happy path by-path and by-fileId still produces canonical filename.

## Test plan
- [x] `go test ./...`: 191 pass
- [x] `go vet ./...`: clean
- [x] `gofmt -l .`: clean
- [x] Integration: 7/7 pass against live dev stack
- [x] Manual: typed 401/403, ambiguous-403 flatten via `errors.As`, happy path unchanged

### DIFF
--- a/downloadclient/errors.go
+++ b/downloadclient/errors.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"mime"
 	"net/http"
+	"unicode/utf8"
 )
 
 // ErrNotSupportedOnV1 indicates the caller requested a feature unavailable
@@ -110,9 +111,12 @@ func formatProblemDetails(status int, pd ProblemDetails, bodySample string) stri
 }
 
 func truncate(s string, n int) string {
-	if len(s) > n {
-		return s[:n] + " [truncated]"
+	if len(s) <= n {
+		return s
+	}
+	for n > 0 && !utf8.RuneStart(s[n]) {
+		n--
 	}
 
-	return s
+	return s[:n] + " [truncated]"
 }
--- a/downloadclient/errors.go
+++ b/downloadclient/errors.go
@@ -1,9 +1,120 @@
 package downloadclient
 
-import "errors"
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"mime"
+	"net/http"
+)
 
 // ErrNotSupportedOnV1 indicates the caller requested a feature unavailable
 // on the v1 download API — either a method with no v1 endpoint
 // (DatasetInfo) or a v2-only option (ListFilesOptions.ExactPath /
 // .PathPrefix).
 var ErrNotSupportedOnV1 = errors.New("operation not supported on v1 download API; use --api-version v2")
+
+// errorBodyReadLimit caps how much of a non-2xx response body is read into
+// memory. 8 KiB is plenty for any realistic RFC 9457 Problem Details
+// document while bounding allocation if a hostile or misconfigured server
+// streams a huge payload.
+const errorBodyReadLimit = 8 << 10
+
+// ProblemDetails is the RFC 9457 Problem Details shape.
+// All fields are optional per the spec.
+type ProblemDetails struct {
+	Type     string `json:"type,omitempty"`
+	Title    string `json:"title,omitempty"`
+	Status   int    `json:"status,omitempty"`
+	Detail   string `json:"detail,omitempty"`
+	Instance string `json:"instance,omitempty"`
+}
+
+// APIError is the typed error returned from a non-2xx HTTP response.
+// StatusCode lets callers discriminate on the response code (e.g. 403
+// handling) via errors.As without substring-matching the message.
+// Problem is populated when the body parses as RFC 9457 Problem Details;
+// nil for non-Problem bodies (HTML error pages, empty bodies, invalid JSON).
+type APIError struct {
+	StatusCode int
+	Problem    *ProblemDetails
+	// Body is the truncated raw response body, bounded for safe display.
+	Body string
+	msg  string
+}
+
+func (e *APIError) Error() string { return e.msg }
+
+// parseErrorResponse converts a non-2xx HTTP response into an *APIError.
+// Tries Problem Details first (Content-Type: application/problem+json
+// or application/json). Falls back to "server returned status N: <body>"
+// when parsing fails or the body isn't Problem Details. Always safe even
+// on empty or malformed bodies.
+//
+// Always consumes and closes resp.Body. The body read is capped at
+// errorBodyReadLimit to bound allocation from a hostile server.
+func parseErrorResponse(resp *http.Response) error {
+	defer resp.Body.Close() //nolint:errcheck
+	body, _ := io.ReadAll(io.LimitReader(resp.Body, errorBodyReadLimit))
+
+	apiErr := &APIError{
+		StatusCode: resp.StatusCode,
+		Body:       truncate(string(body), 200),
+	}
+
+	if isJSONContentType(resp.Header.Get("Content-Type")) && len(body) > 0 {
+		var pd ProblemDetails
+		if err := json.Unmarshal(body, &pd); err == nil {
+			apiErr.Problem = &pd
+			apiErr.msg = formatProblemDetails(resp.StatusCode, pd, apiErr.Body)
+
+			return apiErr
+		}
+		// JSON didn't parse; fall through to the raw-body fallback.
+	}
+
+	apiErr.msg = fmt.Sprintf("server returned status %d: %s", resp.StatusCode, apiErr.Body)
+
+	return apiErr
+}
+
+// isJSONContentType reports whether the media type in ct is
+// application/problem+json or application/json. HTTP media types are
+// case-insensitive per RFC 7231, so we go through mime.ParseMediaType
+// which normalizes the main type to lowercase.
+func isJSONContentType(ct string) bool {
+	mt, _, err := mime.ParseMediaType(ct)
+	if err != nil {
+		return false
+	}
+
+	return mt == "application/problem+json" || mt == "application/json"
+}
+
+// formatProblemDetails picks the best error wording given which
+// RFC 9457 fields are populated. Per the spec all fields are optional,
+// so we handle the no-Title/no-Detail case by surfacing the truncated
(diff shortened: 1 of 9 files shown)


downloadclient/errors.go
```diff
@@ -1,9 +1,120 @@
...
+//
+// Always consumes and closes resp.Body. The body read is capped at
+// errorBodyReadLimit to bound allocation from a hostile server.
+func parseErrorResponse(resp *http.Response) error {
+	defer resp.Body.Close() //nolint:errcheck
+	body, _ := io.ReadAll(io.LimitReader(resp.Body, errorBodyReadLimit))
+
+	apiErr := &APIError{
+		StatusCode: resp.StatusCode,
+		Body:       truncate(string(body), maxErrorBodyBytes),
```
KarlG-nbis, line 64: Why are we truncating the body to 200 bytes?

jhagberg: Good question! There are actually two separate limits here. The `errorBodyReadLimit` (8 KiB) caps how much we read into memory from the response, so a hostile or misconfigured server cannot stream a huge payload at us. The `maxErrorBodyBytes` (200) is a display cap for the error message shown to the user. Without it a full HTML error page or a verbose JSON document would flood the terminal output. So the 8 KiB protects memory, and the 200 bytes keeps the CLI output readable.

The constant was extracted from the earlier PRs in the stack (#681) where zeidlitz flagged the hardcoded `200` as a magic number.

KarlG-nbis: idk imo seems a bit excessive to design this kind of client with it connecting to a "hostile" or misconfigured server in mind, and just leads to additional complexity in the code.

Given we control the server configuration, and we write the instructions for the user on how to connect to the server, I dont see why we should expect a "hostile" or misconfigured server, and if a user wants to connect and use this client and connect to some other endpoint I dont see why we should limit that

jhagberg: Fair point that we control the server. The "hostile" framing was a bit much from my side. But the 200-byte display truncation is actually inherited from the v1 code in `download.go` where it was a hardcoded `200`, we just gave it a name. The more realistic scenario is not a hostile server but a misconfigured reverse proxy returning a full nginx HTML error page on a 502, which would be pretty ugly in the terminal without truncation.

The actual code complexity is just one `io.LimitReader` call and one `truncate()` function, so I think the cost is low for what it gives us. But happy to discuss if you feel strongly about removing it.

KarlG-nbis: Yeah sure makes sense as long is documented / indicated such that client is aware it has been truncated.

But cant find where in the [previous implementation](https://github.com/NBISweden/sda-cli/blob/v0.3.3/download/download.go#L451) you reference when you mentioning the inheritance

jhagberg: yes sorry I my wrong, I looked at my own v1 implementation...

jhagberg: Good point on indicating truncation. What about if I  add a [truncated] suffix when the body gets cut off so the user can tell they are not seeing the full response?

KarlG-nbis (approved):

downloadclient/errors.go
```diff
@@ -1,9 +1,118 @@
...
+		return fmt.Sprintf("%s (%d)", pd.Title, status)
+	case pd.Detail != "":
+		return fmt.Sprintf("server error %d: %s", status, pd.Detail)
+	default:
+		return fmt.Sprintf("server returned status %d: %s", status, bodySample)
+	}
+}
+
+func truncate(s string, n int) string {
+	if len(s) > n {
```
nanjiangshu:
```suggestion
    runes := []rune(s)
	if len(s) > n {
```

downloadclient/errors.go
```diff
@@ -1,9 +1,118 @@
...
+		return s[:n] + " [truncated]"
```
nanjiangshu:
```suggestion
		return string(runes[:n]) + " [truncated]"
```

downloadclient/errors.go
```diff
@@ -1,9 +1,118 @@
...
+	case pd.Title != "":
+		return fmt.Sprintf("%s (%d)", pd.Title, status)
+	case pd.Detail != "":
+		return fmt.Sprintf("server error %d: %s", status, pd.Detail)
+	default:
+		return fmt.Sprintf("server returned status %d: %s", status, bodySample)
+	}
+}
+
+func truncate(s string, n int) string {
```
nanjiangshu, line 113: String slicing operates on raw bytes rather than characters (runes).

If a server error response includes multi-byte UTF-8 characters right around the 200-byte boundary, cutting the string at exactly [:200] can slice a character in the middle of its byte sequence. This will create an invalid UTF-8 string, which will render as a corrupted character.

see below for suggested changes.

nanjiangshu (commented):
Nice work. Added a minor comment regarding potential problems when truncating byte strings.

downloadclient/errors.go
jhagberg: Good catch on the UTF-8 safety. I went with a slightly different approach that keeps the byte-based limit but backs up to the last rune boundary instead of converting to a rune slice:

```go
for n > 0 && !utf8.RuneStart(s[n]) {
    n--
}
```

This way the cap stays at ~200 bytes (which is what we want for terminal output) instead of 200 characters, and avoids the rune slice allocation.

nanjiangshu (approved):

PR #682, opened 2026-04-23
[upload,list] Migrate to feature/s3/transfermanager

jhagberg:
**Related issue(s) and PR(s)**  
This PR closes #658.

**Description**

`feature/s3/manager.Uploader` is deprecated. staticcheck (enabled in `.golangci.yml`) flags 5 `SA1019` on HEAD. Moves `upload/upload.go` and `list/list_test.go` onto `feature/s3/transfermanager`.

- `manager.NewUploader(s3Client)` becomes `transfermanager.New(s3Client, ...)`. `PartSize` moves from a per-call callback onto the client as `PartSizeBytes`. Still `int64`, still read from `helpers.Config.MultipartChunkSizeMb`.
- Dropped `LeavePartsOnError = false`. Old SDK defaulted to `false` anyway, and transfermanager aborts on error by default. No behavioural diff.
- `UploadObjectOutput.Location` is `*string` and only set by the multipart path (`mapFromCompleteMultipartUploadOutput`). Single-part returns `nil`, which broke the 4 upload test assertions. The success message now builds the path-style URL from `HostBase`, `AccessKey`, and the key, `url.PathEscape`'ing each segment so the output matches the old SDK's `EscapedPath()`.
- Adds a 60 MiB random fixture (`data_file_big`) to the integration workflow. With `PartSizeBytes = 50 MiB` the encrypted file is large enough to hit the multipart path. Existing 10 MiB fixture only covered single-part. Upload size is verified against the local file inline.

**How to test**

- `golangci-lint run` → 0 issues (was 5 × `SA1019`).
- `go test ./list/... ./upload/...` → 30 pass.
- `.github/workflows/integration.yml` runs both single-part (10 MiB) and multipart (60 MiB).

PR #689, opened 2026-04-24
fix(upload): opt out of flexible checksum headers for SDA inbox compatibility

jhagberg:
Follow-up to #682 (the transfermanager migration for #658). Fixes a regression spotted in Codex review of that PR.

**Description**
After #682 moved uploads to `feature/s3/transfermanager`, small (single-part) uploads suddenly carried the AWS flexible-checksum headers (`x-amz-sdk-checksum-algorithm: CRC32`, `x-amz-checksum-crc32`, `x-amz-trailer`) — even though the s3 client is built with `RequestChecksumCalculationWhenRequired`, which is supposed to suppress them.

What's going on: `transfermanager.New` runs `resolveChecksumAlgorithm` after applying user options, so an unset `ChecksumAlgorithm` gets rewritten to `CRC32` at construction time. From then on the transfer manager's own value wins over the s3 client setting and ends up on the `PutObjectInput`. The old (now-deprecated) manager path didn't do this, so SDA-style inbox endpoints that don't accept flexible checksums started rejecting normal small uploads after the migration.

The fix is a per-call optFn on `UploadObject` that resets `ChecksumAlgorithm` back to `""`. Per-call optFns run on a copy of the client options and execute before the upload reads the field, so the empty value sticks and no `x-amz-sdk-checksum-algorithm` / `x-amz-checksum-*` / `x-amz-trailer` headers go out. The multipart path reads the same field, so this also clears the headers from `CreateMultipartUpload` and `UploadPart` requests.

**How to test**
- `go test ./upload/... -count=1`. The new test `TestUploadSinglePartDoesNotSendFlexibleChecksumHeaders` wraps the gofakes3 mock with a handler that captures the real PUT request headers, then asserts the three flexible-checksum headers are absent.
- End-to-end: run `sda-cli upload <small-file>` against an SDA inbox that was rejecting uploads after #682 landed, and confirm it now succeeds.

### DIFF
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -229,6 +229,9 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
 			Bucket:          aws.String(config.AccessKey),
 			Key:             aws.String(path.Join(targetDir, outFiles[k])),
 			ContentEncoding: aws.String(config.Encoding),
+		}, func(o *transfermanager.Options) {
+			// Preserve RequestChecksumCalculationWhenRequired for S3-compatible inboxes.
+			o.ChecksumAlgorithm = ""
 		})
 		// Print the progress bar. Second check is to filter out some junk from the output
 		if result != nil && result.VersionID != nil {
--- a/upload/upload.go
+++ b/upload/upload.go
@@ -18,6 +18,7 @@ import (
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager"
+	"github.com/aws/aws-sdk-go-v2/feature/s3/transfermanager/types"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/neicnordic/crypt4gh/keys"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


upload/upload.go
```diff
@@ -229,6 +229,9 @@ func uploadFiles(files, outFiles []string, targetDir string, config *helpers.Con
 			Bucket:          aws.String(config.AccessKey),
 			Key:             aws.String(path.Join(targetDir, outFiles[k])),
 			ContentEncoding: aws.String(config.Encoding),
+		}, func(o *transfermanager.Options) {
+			// Preserve RequestChecksumCalculationWhenRequired for S3-compatible inboxes.
+			o.ChecksumAlgorithm = ""
```
nanjiangshu:
```suggestion
			o.ChecksumAlgorithm = types.ChecksumAlgorithm("")
```

To ensure the field is typed correctly by using a explicit conversion or the empty string variant of the type:

jhagberg: Done in eca7598, pushed. Agreed the explicit conversion reads better — the bare `""` only compiles because Go assigns untyped string constants to named string types automatically, and that's not a rule anyone wants to think about mid-review. Added the `types` import too.

nanjiangshu (approved):
LGTM

jbygdell (approved):

PR #714, opened 2026-08-11
add --bytes flag when listing files

### LINKED ISSUE #709: listing size in bytes does not work
**Describe the bug**
The `sda-cli` should have an option that can list files with the flag that outputs file sizes in bytes instead of human readable sizes

**Steps to reproduce**
Use the -bytes flag and see if it works.

**Expected behavior**
outputs file sizes in byte format

**Additional context**

pahatz:
This PR closes #709.

**Description**
Some users requested the option to be able to list files using the `--bytes` flag so they can compare the uploaded files and make sure the size is correct.

**How to test**
use the local setup, upload a file and list using the flag

### DIFF
--- a/.github/integration/tests/30_list.sh
+++ b/.github/integration/tests/30_list.sh
@@ -9,6 +9,14 @@ else
     exit 1
 fi
 
+size=$(wc -c < data_file.c4gh | awk '{print $1}')
+output=$(./sda-cli --config testing/s3cmd.conf list --bytes)
+if echo "$output" | grep -Eq "^${size}[[:space:]]+data_file\\.c4gh"; then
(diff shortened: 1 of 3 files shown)


KarlG-nbis (approved):

nanjiangshu (approved):
Look good to me.

However, the PR description is a bit misleading, the flag `--bytes` already exist, just the format might not be ideal to some users. 

For `sda-cli-v0.3.3` it shows 

e.g. `164.00B`, the user probably just want `164`, right?

nanjiangshu:
Hi @pahatz , I think it's ready to be merged, what do you think?

